PR #305, opened 2019-08-17
Fix incorrect French/Français label

### LINKED ISSUE #289: Wrong term "French" instead of "Français" in language menu
In the language drop down of the menu the French language is labeled as "French".
It should be replaced by "Français".

danielthegray:
Fixes #289


mdoelker:
Good catch

PR #311, opened 2019-09-02
Beta/accra

RaimundWege:


lib/blocs/request_manager/online_request_manager.dart
```diff
@@ -99,7 +99,7 @@ class OnlineRequestManager implements RequestManager {
       .replace(queryParameters: {
         "fromPlace": from.toString(),
         "toPlace": to.toString(),
-        "date": "01-01-2018",
+        "date": "04.03.2019",
```
mdoelker: Can we make this the current date in general?

lib/trufi_models.dart
```diff
@@ -156,7 +156,7 @@ class TrufiStreetJunction {
   final double longitude;
 
   String get description {
-    return "${street1.location.description} y ${street2.location.description}";
+    return "${street1.location.description} and ${street2.location.description}";
```
mdoelker: Uh, we need to put this into translation asap :)

lib/blocs/request_manager/online_request_manager.dart
RaimundWege: Is this also a question for the backend team? At the beginning we had to specify this date as a fixed date because we prepared only one day in our gtfs file.

mdoelker: Ahh okay, I was not aware of that.

mdoelker: Don't think it's necessary though. We currently specify the timetable as valid from year 2000-2100. And if there would be a change we could control this from the backend. So the client should send the date it is searching for, which in our case would always be today for now.

aemkei: Right. We set the date to a fixed day, because the first version of our GTFS was limited to a specific date. If we you changed that to a valid timetable, it should be fine

PR #312, opened 2019-09-03
Moved map settings to config

mdoelker:


assets/cfg/app_config.json.dist
```diff
@@ -10,6 +10,17 @@
...
     "keyMapTiler": "",
+    "mapDefaultZoom": 12.0,
+    "mapOfflineMinZoom": 8.0,
+    "mapOfflineMaxZoom": 14.0,
+    "mapOfflineZoom": 13.0,
+    "mapOnlineMinZoom": 1.0,
+    "mapOnlineMaxZoom": 19.0,
+    "mapOnlineZoom": 13.0,
+    "mapChooseLocationZoom": 16.0,
+    "mapCenterCoords": [-17.39000, -66.15400],
```
amenk: I think we should make clear here, what's lat and what's lon ...

mdoelker: Good point

Issue #315, opened 2019-09-28
Query date is hard coded to 01-01-2018, use the current date instead


amenk:
While trying to connect to an own otp server I figured out, that now plans were found. Turns out the query date is hard coded:

https://github.com/trufi-association/trufi-app/blob/6d3d774325c0f055430ca4f8a38c6df74102cfd1/lib/blocs/request_manager/online_request_manager.dart#L102

We should use the current day here, right?

I can make a pull request.

Own note:

The date format in the query is MM-DD-YYYY (

        "date": "09-28-2019",

worked for my use case very well, and now I see results.

Actually i found https://github.com/trufi-association/trufi-app/issues/2 where this was intentionally set. But can't just the OTP server be fixed and provide some (fake) schedules which are just always valid? Seems like a much cleaner solution and you do not have to update the app, when a real schedule is published / changed.

mdoelker:
We have talked about this internally before, and we should use the current date here. We might have real schedules in the future, so we would need this anyway. I'm not sure what the issue was exactly that we chose this date. It should work on all days really, as we are setting up the GTFS so that the generated trips are valued mo-su throughout the year way into the future.

@aemkei can you shine some light onto this?

mdoelker:
https://github.com/trufi-association/trufi-app/pull/311 Discussion can be found here for reference

PR #316, opened 2019-09-29
#315 Use today as query date

amenk:
Fetch the current date and format in Month-Dy-Year Order for the API
request.

No using intl lib, because it would cause overhead, and we have a fixed
format here.

Before this fix, custom GTFS server which only have a recently dated
schedule would not return results.


mdoelker:
Resolves #315

PR #317, opened 2019-09-29
Fix/313 readme: Improve readme

amenk:
* Copy-paste enable cp command
* Add note about keystore


mdoelker:
Great addition. Thanks!

PR #322, opened 2019-10-03
Feat/share

### LINKED ISSUE #321: Introduce a "Share this app" button

mdoelker:
App share menu item. Resolves #321

### DIFF
--- a/assets/cfg/app_config.json.dist
+++ b/assets/cfg/app_config.json.dist
@@ -10,6 +10,7 @@
     "urlTwitter": "",
     "urlInstagram": "",
     "urlDonate": "",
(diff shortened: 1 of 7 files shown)


mdoelker:
Based on #319, merge first

amenk:
This seems to break other-than en_US translations with a red-screen-of-death :-(
As soon as you try to access the burger menu "Unsupported operation: No translation found".
Would be good to have a CI-job which tests that all languages have all translations.

mdoelker:
The more languages we support the more unlikely it will be that we always have all translations, so we also need a way to handle a miss. See my comment in #328

PR #338, opened 2019-10-14
Transform app to plugin **WIP**

RaimundWege:
This branch is work in progress and currently just for discussion. We're trying to transform the whole app into a plugin which allows us to create new projects and import our trufi logic just by a single dependency.


pubspec.yaml
```diff
@@ -1,23 +1,24 @@
-name: trufi_app
-description: Trufi App
+name: core
```
mdoelker: Is this really supposed to be named core or rather trufi core?

RaimundWege: I've tested also `trufi_core` but then it generates `app.trufi.trufi_core` as package name instead of `app.trufi.core`. I want to change the name back to `trufi_core` but keep `app.trufi.core` as package name in the next step.

RaimundWege: From now on the name of the plugin is `trufi_core`.

pubspec.yaml
```diff
@@ -42,51 +43,34 @@ dev_dependencies:
...
+  #
+  # For details regarding assets in packages, see
+  # https://flutter.dev/assets-and-images/#from-packages
+  #
+  # An image asset can refer to one or more resolution-specific "variants", see
+  # https://flutter.dev/assets-and-images/#resolution-aware.
   assets:
-    - assets/cfg/
     - assets/data/
     - assets/images/
```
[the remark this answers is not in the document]
RaimundWege: Correct. Search data has already been removed. But we need some images as fallback if the user has no images to start with.

pubspec.yaml
```diff
@@ -42,51 +43,34 @@ dev_dependencies:
...
+# following page: https://dart.dev/tools/pub/pubspec
 
 # The following section is specific to Flutter.
 flutter:
+  # This section identifies this Flutter project as a plugin project.
+  # The androidPackage and pluginClass identifiers should not ordinarily
+  # be modified. They are used by the tooling to maintain consistency when
+  # adding or updating assets for this project.
+  plugin:
+    androidPackage: app.trufi.core
```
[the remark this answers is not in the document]
RaimundWege: This code were autogenerated by Flutter. In iOS you don't need a package namespace like this. But you have to set your Bundle ID and Development Team ID in the project settings in Xcode.

Issue #343, opened 2019-11-17
Add bus operator to route instructions


amenk:
In Addis Ababa, we have different bus operators. It is important to see in the time schedule, which is the bus operator. The information is in agencyName in the field of the leg in the OTP response.

Should this be implemented in trufi_core or in the specific Addis Ababa app?

christophhanser:
I would vote for Addis Abeba only since this requirement has not yet popped
up in the other cities.

amenk:
@christophhanser Technically we still might need to implement it as an option in core that is switched of by default, I think.

@RaimundWege or did you plan another simple way to add such functionality to the custom app, which is based on the Trufi core package?

christophhanser:
Now that you say it, I guess that this is the Technically correct approach.
:)

amenk:
I will try to implement that --- we might not need an option, as api.trufi.app does not supply agency info, so I would just not display it, if the info is not there :)

amenk:
tried to implement it (was pretty easy)... but it turns out that for some (bus) it makes sense to show the operator, for others (minibus) not. We will maybe just add the agency name to the route. I am closing this for now.

amenk:
What we of course could do is, to define a schema for agency name, like "Offical Agency Name [Shortname]" and show the text from the brackets - if present - in the itinerary.

md-weber:
Is this issue still relevant for the Core? @christophhanser @amenk

ValorNaram:
Closing this due to topic inactivity.

PR #344, opened 2019-11-17
#343 Support displaying short agency name on itinerary

amenk:


lib/trufi_models.dart
```diff
@@ -366,6 +366,7 @@ class PlanItineraryLeg {
     this.distance,
     this.duration,
     this.toName,
+    this.agencyName
```
RaimundWege:
```suggestion
    this.agencyName,
```
We leave an extra `,` at the end for automatic code formatting.

lib/trufi_models.dart
```diff
@@ -421,7 +425,8 @@ class PlanItineraryLeg {
       _RouteLongName: routeLongName,
       _Distance: distance,
       _Duration: duration,
-      _To: {_Name: toName}
+      _To: {_Name: toName},
+      _AgencyName: agencyName
```
RaimundWege:
```suggestion
      _AgencyName: agencyName,
```

amenk:
I might revert this as we changed our strategy for Addis Ababa and might use a single agency GTFS, so we might put the bus type in the route name

mdoelker:
Does this still apply?

amenk:
We don't use this anymore

PR #351, opened 2019-12-27
No trufi icon in app drawer

amenk:
The Trufi icon should not be used here, because it makes derived apps
harder / impossible to customize properly.

Also the Trufi icon does not make much sense here, as it is a web page
link and a web symbol should be used.

### DIFF
--- a/lib/widgets/trufi_drawer.dart
+++ b/lib/widgets/trufi_drawer.dart
@@ -112,7 +112,7 @@ class TrufiDrawerState extends State<TrufiDrawer> {
           Divider(),
           if (cfg.url.website != "")
             _buildWebLinkItem(
-              CustomIcons.trufi,
+              Icons.web,
               localization.readOurBlog(),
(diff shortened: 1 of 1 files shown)


mdoelker:
I think the `open_in_browser` or `open_in_new` icons probably convey what you are trying to do here better then the `web` icon. But we use the Trufi icon here in the same sense that we also use the FB and Twitter icon, even though they open a web page.

I agree though that this should be configurable in some way.

amenk:
okay, the next question would be to document / link to a documentation how those custom fonts were created :)

mdoelker:
Have a look here https://github.com/trufi-association/trufi-app/blob/feat/293-generalize-code/fonts/CustomIcons/README.md

amenk:
Should I try to make it configurable then?

mdoelker:
Yeah, I think we can make it a configuration parameter for now until we have a better solution. Feel free to implement that.

PR #352, opened 2019-12-27
The bus stop icon was changed for a icon more descriptive

GustavoTCh:
more descriptive one.


mdoelker:
Thanks for your input! I picked up your idea but used a custom bus stop icon in 6ae6e82aaf94a46c383dfd728425ebbe2c7c6517

Issue #265, opened 2019-02-14
Add advertising


hoodi:
We could add ads to the app but this needs to be discussed.

amenk:
One option would be to include a webview with a URL like `https://placeinfo.trufi.app/?lat=...&lon=...` which displays information about the target of a trip, including local businesses

amenk:
What do you think about this form ob subtle ad placement (I think a webview would make it too obtrusive) 

![Selection_252](https://user-images.githubusercontent.com/1087128/71533710-a72e8400-28fa-11ea-821e-a63f2cc41b43.png)

amenk:
The ad server could be as simple as returning a json
    { text: "(Ad)

amenk:
> something like a device ID (to not always display the same ads -> but for data privacy reasons of
> the users a random installation ID would be best)

We should use the ne correlation ID :-)

https://github.com/trufi-association/trufi-app/commit/623ff5fd5d81048606fa7dce0a867a7d8a4d152a#diff-c04dfa7fa351a6a9c77cf2655a39bcb4R109

PR #353, opened 2019-12-28
#265 Add advertising

amenk:
* Add URL for ad server, which returns a simple JSON like this

      {
        "ad": {
          "text": "(Ad) Trufi is awesome.",
          "location": {
            "name": "Trufi Association",
            "lat": 1.23,
            "lon": 4.56
          }
        }
      }

* If the request fails for any resons, it gracefully proceeds and just
displays no ad

* The ad is displayed below the itinerary and should be matching the
destination


mdoelker:
Closed in favor of #364

PR #362, opened 2020-01-24
Feat/share locations

GustavoTCh:
**Feature: share locations(ANDROID)**
It's only for android, not how much with the tools to perform the feature for iOS  :D

Request in ANDROID:
The uni_links package needs to modify the AndroidManifest file.

```
<intent-filter android:autoVerify="true">
              <action android:name="android.intent.action.VIEW" />
              <category android:name="android.intent.category.DEFAULT" />
              <category android:name="android.intent.category.BROWSABLE" />
              <data
                android:scheme="https"
                android:host="trufi.app" />

```
he scheme needs to be "https".
The host may be different, it depends on what types of url your app can open.

The app supports multiple instances, default property of activity is 

`android:launchMode="singleTop"`

LINKS:
 **-Link route  https://trufi.app?type=route&origin=Atomic&lngo=-0.2659963&lato=5.527639&destiny=Challenge&lngd=-0.209728&latd=5.576912**
 **-Link location https://trufi.app?type=location&description=Atomic&lng=-0.2659963&lat=5.527639**

Link's query is very important.

This feature has 2 parts, **export locations-routes** and **import locations-routes**.

**1) Export locations-routes:**

**Export-Import location**

**-Exporting the location from the search screen**
Screenshot_20200123-121358

**-Select the app to share, this case is Whatsapp**
Screenshot_20200123-121407_Android System

Screenshot_20200123-121418_WhatsApp

**OPEN LINK**
Screenshot_20200123-121300_Android System

**Show options, this example is in Trotro, then I have a error because I am not in Accre**
Screenshot_20200123-121436

Screenshot_20200123-121511

**Select save location**
Screenshot_20200124-123141

**Export-Import route**

**When a route is displayed on the screen, a floatingActionButton appears to share the screen route**
Screenshot_20200123-121129

**Select the app to share, this case is Whatsapp**
Screenshot_20200123-121154_Android System

Screenshot_20200123-121231_WhatsApp

**OPEN LINK ROUTE**

Screenshot_20200123-121323


mdoelker:
First of all, thank you for the PR and sorry for not looking at this in more detail earlier. The main reason we did not tend to this immediately is that I wanted to have a consistent strategy for universal links before we release it to the public. The problem here is that when the receiver of the share link does not have the app installed, he will be dropped on the trufi app promotional page without further instructions. Ideally, the shared URL would show some fallback information instead, like a map preview and links to the app stores. That would even work when opening the URL from a desktop device.

Now that we pick up some speed again in the dev team, I'll think about how to solve this so we can get this feature merged. Also, we should have a look on the iOS side of this feature before.

On a technical note, this needs a cleanup before it can be merged because it still contains changes from the saved places feature. We also need to make the URL prefix configurable to work with the different apps. A constant won't work.

christophhanser:
Are these Tasks then todo?
- Clarify a good webpage link that lets the User download the app
- Clarify how to do it with iOS
- Use app Domain like trufi.app flexible within URL

mdoelker:
@christophhanser No, your tasks are misleading.

I would not say it needs clarification, we just have to do it. We can start with a simple fallback page that explains what kind of link this is and that you need the app for it to work. But to be truly universal in the original sense, we should provide a way to at least see what route was shared so that you can see it even on a platform where trufi is not supported (e.g. your laptop). The idea with universal links is that they are just normal website links with their normal content. But then an app can register to handle that URL to provide _the same_ content inside of the app instead. If you get a completely different experience depending on what kind of device you use (or even a 404) you are doing it wrong. Besides, you have to be careful to not completely break the website. The way the share link is designed right now would make the trufi.app website unusable from the device where the app is installed, because the app would activate for all trufi.app URLs.

I would also say it is clear how to do it for iOS. There is plenty of documentation in the uni_links plugin and on Apple's sites. I think it's rather that Gustavo is not familiar with the platform and has no access to a device to test it. The plugin abstracts all the nitty-gritty details for the in-app part. We then just have to host a file on the webserver to allow the app to claim the URL namespace.

The URL could look like https://www.trufi.app/share/location?..., https://www.trufi.app/share/route?... or something like that. It should be configurable though. And not just on the domain level, but the full prefix. We can not just assume that our choice for the URL works for everybody.

ValorNaram:
> The URL could look like https://www.trufi.app/share/location?..., https://www.trufi.app/share/route?... or something like that. It should be configurable though. And not just on the domain level, but the full prefix. We cannot just assume that our choice for the URL works for everybody. You also have to be careful to not completely break the website. The way the share link is designed right now would make the trufi.app website unusable from the device where the app is installed, because the app would activate for all trufi.app URLs.

But not for subdomain.trufi.app because it is another url/domain. So we could set up <subdomain>.trufi.app e.g. `share.trufi.app` and then `share.trufi.app/location?` or `share.trufi.app/route?`.

It would/should then technically no problem.

mdoelker:
`https://www.trufi.app/share/location?` or `https://share.trufi.app/location?` or `https://something.else.completely.com/location?`... doesn't really matter and all works. The point is, it should so something meaningful in cases where it is not handled by the app.

ValorNaram:
> The point is, it should do something meaningful in cases where it is not handled by the app

http(s:)//share.trufi.app/location? could show the location of the user who has shared this link. The geo data could be encoded in the query string of the url.

> share.trufi.app/route?

Each route has an unique id in OSM. We could download route data from OSM or own servers and then display it to the user in the design of the app.

mdoelker:
Closing this for now as it is stale. We need to revisit this later when we have the capacity to build a complete experience around it. I definitely want to get this in eventually. Thank you for your contribution.

PR #387, opened 2020-10-29
Feat/bike planning

mdoelker:


md-weber:
Is this PR still valid?

lib/trufi_models.dart
```diff
@@ -443,20 +442,44 @@ class PlanItineraryLeg {
   }
 
   String _carTypeString(TrufiLocalization localization) {
```
md-weber, line 444: I think we should rename this method, it seems that is more the transportation type.

lib/trufi_models.dart
```diff
@@ -443,20 +442,44 @@ class PlanItineraryLeg {
...
-            : carType.contains(type_micro)
-                ? localization.instructionVehicleMicro()
-                : carType.contains(type_minibus)
-                    ? localization.instructionVehicleMinibus()
-                    : carType.contains(type_gondola)
-                        ? localization.instructionVehicleGondola()
-                        : carType.contains(type_light_rail)
-                            ? localization.instructionVehicleLightRail()
-                            : localization.instructionVehicleBus();
+    switch (mode) {
```
md-weber, line 445: It would be amazing in the future if we can somehow create an Enum out of the different strings. I took a brief look but I believe we first need some kind of Repo Layer with a good mapper.

And something like a JSON Serializer

lib/trufi_map_utils.dart
```diff
@@ -14,21 +14,25 @@ import './trufi_configuration.dart';
...
     urlTemplate: "assets/tiles/{z}/{z}-{x}-{y}.png",
   );
 }
 
-tileHostingTileLayerOptions(String tilesEndpoint, { String tileProviderKey = "" }) {
-  var urlTemplate = tilesEndpoint + "/{z}/{x}/{y}@2x.png";
+tileHostingTileLayerOptions(
+  String tilesEndpoint, {
+  String fileFormat = "png",
+  String tileProviderKey = "",
```
md-weber, line 27: Maybe it would be good to give a verbose warning here if there is no tileProviderKey. 
> Warning: There is no ProviderKey set for the TileProvider.

Therefore, we would make it more obvious what is going on.

lib/trufi_localizations.dart
```diff
@@ -946,6 +976,16 @@ abstract class TrufiLocalization {
 
   String instructionVehicleLightRail();
 
+  String instructionVehicleTrain();
```
md-weber, line 979: I believe these translations are not inside of Lokalise, it would be great to add them.

lib/trufi_configuration.dart
```diff
@@ -92,3 +93,8 @@ class TrufiConfigurationUrl {
   var donate = "";
   var share = "";
 }
+
+class TrufiConfigurationFeatures {
```
md-weber, line 97: To make the configurations more speakable, I would recommend (also for my PR) adding documentation on top of it.

/// [showCarRouteOnTransitError] does ...

lib/pages/home.dart
```diff
@@ -414,27 +415,35 @@ class HomePageState extends State<HomePage>
...
+            context,
+            _data.fromPlace,
+            _data.toPlace,
+          );
+        }
 
         Plan plan = await _currentFetchPlanOperation.valueOrCancellation(null);
 
         if (plan == null) {
           throw "Canceled by user";
```
md-weber, line 441: This is not part of this PR, but we should create a fitting Error class for something like that in the future.

lib/blocs/request_manager/online_request_manager.dart
```diff
@@ -155,14 +164,13 @@ class OnlineRequestManager implements RequestManager {
     if (response.statusCode == 200) {
       return await compute(_parseAd, utf8.decode(response.bodyBytes));
     } else if (response.statusCode == 404) {
-      print ("No ads found");
+      print("No ads found");
```
md-weber, line 167: Instead of using print, I would recommend using `developer.log`
https://flutter.dev/docs/testing/code-debugging#logging

lib/blocs/request_manager/online_request_manager.dart
```diff
@@ -110,22 +117,24 @@ class OnlineRequestManager implements RequestManager {
...
+      "date": _todayOtpDate(),
+      "time": _todayOtpTime(),
       "numItineraries": "5",
       "mode": mode,
       "correlation": preferences.correlationId,
-    });
+    };
+    if (additionalParams != null) {
+      params.addAll(additionalParams);
+    }
```
md-weber, line 134:
```dart
    final params = {
      "fromPlace": from.toString(),
      "toPlace": to.toString(),
      "date": _todayOtpDate(),
      "time": _todayOtpTime(),
      "numItineraries": "5",
      "mode": mode,
      "correlation": preferences.correlationId,
      if (additionalParams != null) ...additionalParams
    };
```

lib/blocs/request_manager/offline_request_manager.dart
```diff
@@ -66,6 +66,18 @@ class OfflineRequestManager implements RequestManager {
     }());
   }
 
+  CancelableOperation<Plan> fetchBikePlan(
+    BuildContext context,
+    TrufiLocation from,
+    TrufiLocation to,
+  ) {
+    return CancelableOperation.fromFuture(() async {
+      throw FetchOfflineRequestException(
```
md-weber, line 75: Would it not be better to throw the correct type
```suggestion
      throw UnimplementedError("Fetch plan as bike route offline is not implemented yet.");
```

md-weber:
Some tests would be amazing :)

lib/trufi_models.dart
mdoelker: Some of these types are also very specific to the country/region the app is used in. We should probably introduce a way to extend the list from the host app and move things like Microbus/Trufi to it. The core would still know how to translate generic trains, trams, buses etc.

lib/blocs/request_manager/online_request_manager.dart
mdoelker: Haha, true. IIRC that feature was introduced to Dart after this code was written :)

lib/blocs/request_manager/offline_request_manager.dart
```diff
@@ -66,6 +66,18 @@ class OfflineRequestManager implements RequestManager {
...
   }
 
+  CancelableOperation<Plan> fetchBikePlan(
+    BuildContext context,
+    TrufiLocation from,
+    TrufiLocation to,
+  ) {
+    return CancelableOperation.fromFuture(() async {
+      throw UnimplementedError("Fetch plan as bike route offline is not implemented yet.");
+        Exception("Fetch plan as bike route offline is not implemented yet."),
```
md-weber, line 76:
```suggestion

```

lib/trufi_map_utils.dart
mdoelker: A key may not be required depending on the tile provider, so a warning would not make too much sense :)

mdoelker:
Main changes were re-implemented in current development branch

PR #390, opened 2020-11-23
Fix/from marker center

SamuelRioTz:
Fix from marker center and upgrade **geolocator** dependency 
WhatsApp Image 2020-11-23 at 05 18 17
WhatsApp Image 2020-11-23 at 05 18 16

### DIFF
--- a/lib/blocs/location_provider_bloc.dart
+++ b/lib/blocs/location_provider_bloc.dart
@@ -50,8 +50,9 @@ class LocationProviderBloc implements BlocBase {
   Future<LatLng> get currentLocation async {
     LatLng location;
     try {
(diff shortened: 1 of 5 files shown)


lib/blocs/location_provider_bloc.dart
```diff
@@ -77,22 +78,22 @@ class LocationProvider {
 
   final CompositeSubscription _subscriptions = CompositeSubscription();
   final Geolocator _geolocator = Geolocator();
```
mdoelker, line 80: I think this can be removed now
```suggestion
```

mdoelker (approved):

PR #393, opened 2021-01-26
Feature/custom widget

SamuelRioTz:
adding support for custom widgets for home screen
Screen Shot 2021-01-26 at 01 00 02
Screen Shot 2021-01-26 at 01 13 12


mdoelker:
Code looks good, but could we improve the naming, so it is clear where this custom widget is rendered? In case we add more later we need this to be more specific.

SamuelRioTz:
agree. 
we also may need some extra container, what name should I use
Screen Shot 2021-01-26 at 10 31 40

mdoelker:
hmm maybe `mapOverlayWidget`, `mapExtraFabWidget` (as in Floating Action Button)

md-weber:
This is how the app behaves at the moment.

grafik

lib/pages/home.dart
```diff
@@ -30,6 +29,14 @@ import '../widgets/trufi_drawer.dart';
 
 class HomePage extends StatefulWidget {
   static const String route = '/';
+  final LocaleWidgetBuilder customOverlayWidget;
+  final WidgetBuilder customBetweenFabWidget;
+
+  const HomePage(
+      {Key key,
+      @required this.customOverlayWidget,
```
md-weber, line 37: I am not sure if they should be required or optional.

lib/pages/home.dart
```diff
@@ -30,6 +29,14 @@ import '../widgets/trufi_drawer.dart';
...
+      @required this.customBetweenFabWidget})
```
md-weber, line 38: I am not sure if they should be required or optional.

lib/plan/plan_empty.dart
```diff
@@ -37,16 +49,39 @@ class PlanEmptyPageState extends State<PlanEmptyPage>
...
         right: 16.0,
         child: _buildLowerActionButtons(context),
       ),
+      Positioned.fill(
+        child: Container(
+          margin: EdgeInsets.only(
+            right: customOverlayWidgetMargin,
+            bottom: 60,
+          ),
+          child: widget.customOverlayWidget != null
```
md-weber, line 68: We even check against null, so it would make sense to not make them required.

lib/plan/plan_map.dart
```diff
@@ -99,16 +109,40 @@ class PlanMapPageState extends State<PlanMapPage>
...
+        ),
+        Positioned.fill(
+          child: Container(
+            margin: EdgeInsets.only(
+              left:
+                  MediaQuery.of(context).size.width - customOverlayWidgetMargin,
+              bottom: 80,
+              top: 65,
+            ),
+            child: widget.customBetweenFabWidget != null
```
md-weber, line 141: Same here

md-weber (commented):
I think it is good, it would be great if we remove the required fields for the widgets.

lib/pages/home.dart
mdoelker: Optional

lib/pages/home.dart
mdoelker: Optional

md-weber (approved):

Issue #330, opened 2019-10-08
Migrate from Google Translate Toolkit


mdoelker:
Google Translate Toolkit is shutting down and it is the only platform supporting dart intl's ARB format. Track progress in this issue https://github.com/dart-lang/intl/issues/230 and come up with a migration plan to a different platform.

This might help https://github.com/bmw-tech/arb-converter-cli

christophhanser:
Sorry for asking maybe stupid but would Localization Lab help?

mdoelker:
Published tool to convert to XLIFF and back https://www.npmjs.com/package/arb-convert

mdoelker:
Working on extending our bash scripts to automate import/export from our new localization provider.

md-weber:
With Flutter 2.0 we changed the way internationalization works entirely in the Trufi Core. Now we are using the default way gen-l10n that got introduced by Flutter 1.22. 

I created a Script to get the translations directly from Lokalise via the [Lokalise2 CLI](https://github.com/lokalise/lokalise-cli-2-go). In the future this will be a one-way process, we will add translated Strings to our Lokalise project and can execute the script to receive the translations in our project.

I created more information in our Wiki. 
- [How to work with custom Translations](https://github.com/trufi-association/trufi-core/wiki/Custom-Translations)
- [How to update the strings from Lokalise](https://github.com/trufi-association/trufi-core/wiki/Translations-Update-Guide)

Issue #292, opened 2019-07-18
Language files need to be generalized


christophhanser:
"title": "Trufi App" => the title is unique across all languages, but different for each app, so we should have it saved some where else

"tag_line": "Public transportation in Cochabamba" => city should be removed here and made generic, so that Accra and other cities work as well

mdoelker:
I'm not sure if we should pull this out of the translation files. Maybe it makes sense to externalize translations out of the core package that we want to build (#293)

amenk:
Why not putting the cites as a variable? So most of the translations can be reused ...

md-weber:
Without knowing this Issue I did exactly that, in Lokalise I added the cityName as a variable and it is now possible to pass that down. Thanks to the TrufiCustomTranslations it is now also possible to overwrite the Title.

PR #413, opened 2021-04-10
👌 [Quality Improvement] Introduction Flutter 2.0 + Translation Sync

### LINKED ISSUE #397: Upgrade to Flutter 2.0
We currently have many old dependencies, and it would be great to bring the application to Flutter 2.0.

md-weber:
# Introduction Flutter 2.0
Every package was easy to upgrade to Flutter 2.0 only fight was the intl_translation package. I replaced it with the gen-l10n approach that Flutters introduced in 1.22.

It is a quite large PR, but it improved all packages, and the app runs on my part. It would be amazing to check out this PR and see if it works with the correct Server.

Unfortunatly there are just a minor set of tests to make sure that the application is still running consistently. 

## Open Questions - Discussions
- Should we commit the l10n files, or should every developer generate them locally? ✔
  - At the moment, we generated them and provided them in the Repository

- Host App can no longer pass down easily a "TrufiTranslation" because it is generated inside the Core. ✔
  - It can still pass down new information via the TrufiConfiguration
  - There is now a custom translation object which takes for several specified Translations the value
  - If the value of the Map is null, it displayed the default Localization
  - If the Local contains a LangKey and CountryKey and the CountryKey does not exist, it falls back to the only LangKey, and if it does not exist, it falls back to the default Local.

## Example Translation (Trotro App)
```dart
void setupTrotroTranslations(TrufiConfiguration config) {
  config.customTranslations
    ..title = {
      Locale("de"): "Trotro App",
      Locale("en"): "Trotro App",
      Locale("es"): "Trotro App",
      Locale("fr"): "Trotro App",
      Locale("it"): "Trotro App",
      Locale("qu"): "Trotro App",
    };
}
```

## Migration Path
Because of Flutter 2.0, we have several breaking changes for the dependencies. I would recommend that we upgrade the Trufi Core and change the child's dependencies to a commit/release the past. This should be a release of Trufi Core 2.0.

https://github.com/trufi-association/trufi-core/commit/3d054344135325fe749072a47605509aeb64a38c

I created a brief Draft Release https://github.com/trufi-association/trufi-core/releases/tag/untagged-4378e1fb2b0d71d90a6e

## Documentation
In order to make it easier to use for our End Customers, I added a Wiki into our project.
https://github.com/trufi-association/trufi-core/wiki/Custom-Translations
https://github.com/trufi-association/trufi-core/wiki/Translations-Update-Guide

Resolves #397 #330 #292


lib/location/location_form_field.dart
```diff
@@ -24,12 +23,6 @@ class LocationFormField extends FormField<TrufiLocation> {
               padding: EdgeInsets.all(4.0),
               child: GestureDetector(
                 onTap: () async {
-                  // Customize search hint
```
md-weber, line 27: This part was rather interesting, it sets a "HintText" but the hint text is never read via the getter. Therefore this was deprecated and I removed it instead of repairing it.

lib/trufi_app.dart
```diff
@@ -145,14 +144,10 @@ class _LocalizedMaterialAppState extends State<LocalizedMaterialApp> {
             );
           },
           localizationsDelegates: [
-            TrufiLocalizationsDelegate(
-              snapshot.data,
-              widget.localization,
```
md-weber, line 150: This is the only area I am unsure of, it seems that we were able to overwrite the location from the host system. This would not be possible anymore and they needed to create the arb file and generate it inside of Trufi_Core.

pubspec.yaml
```diff
@@ -13,35 +13,35 @@ dependencies:
...
+  synchronized: ^3.0.0
+  url_launcher: 6.0.3
   uuid: 2.2.2
 
 dev_dependencies:
   flutter_driver:
     sdk: flutter
   flutter_test:
     sdk: flutter
+  network_image_mock: 1.1.0
```
md-weber, line 44: This new dependency comes from https://github.com/stelynx/network_image_mock. It allows us to test the package. The benefit is already ready for Flutter Null Safety.

lib/trufi_app.dart
mdoelker: We discussed that we provide custom translations for specific keys to the host app.

lib/widgets/trufi_drawer.dart
```diff
@@ -26,12 +26,14 @@ class TrufiDrawer extends StatefulWidget {
...
-  final GlobalKey appShareButtonKey = GlobalKey(debugLabel: "appShareButtonKey");
+  final GlobalKey appShareButtonKey =
+      GlobalKey(debugLabel: "appShareButtonKey");
 
   @override
   void initState() {
     super.initState();
 
+    // TODO: Should have some kind of fallback image
     bgImage = AssetImage(TrufiConfiguration().image.drawerBackground);
```
mdoelker, line 37: I suggest we don't set an image then. It will have the primary color as the background then.

mdoelker:
`--untranslated-messages-file=desiredFileName.txt` in the translation conversion scripts is probably an oversight :)

mdoelker:
Code looks good, haven't built and tested so far though.

lib/location/location_form_field.dart
mdoelker: Yeah, I think we can remove it then. Or is it obvious where the hint should have been displayed? The param would be obsolete as well.

lib/widgets/trufi_drawer.dart
md-weber: Yes, I thought the same, but for the moment, I did not want to change any more logic. It would be a separate PR where I go through all the TODO´s I created.
I could also change the Todo text if you like.

lib/location/location_form_field.dart
md-weber: I cross-checked the Form Fields that I could see. Nothing really needed that flag. Also, I removed now the parameter, also no obvious behaviour change. Of course, without tests, it is hard to say that there is no Side effect somewhere.

lib/blocs/bloc_provider.dart
```diff
@@ -18,12 +18,10 @@ class BlocProvider<T extends BlocBase> extends StatefulWidget {
   _BlocProviderState<T> createState() => _BlocProviderState<T>();
 
   static T of<T extends BlocBase>(BuildContext context) {
-    final type = _typeOf<BlocProvider<T>>();
-    BlocProvider<T> provider = context.ancestorWidgetOfExactType(type);
+    BlocProvider<T> provider =
+        context.findAncestorWidgetOfExactType<BlocProvider<T>>();
```
SamuelRioTz, line 22: does this change not break something?

md-weber: For me, it did not change anything. Unfortunatly the PR is again broken because I try to sync Lokalise with the PR. It's quite a mess. But after that, it would be great if you could test it.

md-weber: 🎉Should work again.

mdoelker (approved):

Issue #402, opened 2021-04-10
Code Analysis with lint package


md-weber:
The code analyzer allows us to reduce bad patterns in code and improves our codebase a lot. Therefore, I would recommend adding a code linter to our project.

There is a fantastic Lint package and allows us to analyze for anti-pattern like not using the const keyword if possible. Just a heads-up, an introduction of a Linter usually leads to a large initial PR. Later this should be part of our build pipeline.

https://pub.dev/packages/lint

PR #419, opened 2021-04-18
🧹 [Linting] Added the Lint package to Trufi - Core

md-weber:
I introduced the Lint package into our codebase. Unfortunatly there is currently an open Issue in Dart that prevents us from "just" using the package I needed to copy the analysis_options.yaml over to make it work for our project.

The issue to use the Lint package directly: https://github.com/dart-lang/sdk/issues/42910

Resolves #https://github.com/trufi-association/trufi-core/issues/402

For future references I think we really should stop catching errors and only print some text.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -5,6 +5,8 @@ import 'package:latlong/latlong.dart';
 import 'package:trufi_core/trufi_app.dart';
 import 'package:trufi_core/trufi_configuration.dart';
 
+// TODO: Remove the async GlobalConfig load
+// https://github.com/trufi-association/trufi-core/pull/419#discussion_r615767538
 Future<void> main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -14,6 +14,8 @@ Future<void> main() async {
   final globalCfg = GlobalConfiguration();
   await globalCfg.loadFromAsset("app_config");
 
+  trufiCfg.map.mapTilerKey = globalCfg.getValue<String>("mapTilerKey");
+
   // Abbreviations
   trufiCfg.abbreviations.addAll({
     "Avenida": "Av.",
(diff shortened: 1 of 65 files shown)


example/lib/main.dart
```diff
@@ -5,7 +5,7 @@ import 'package:latlong/latlong.dart';
 import 'package:trufi_core/trufi_app.dart';
 import 'package:trufi_core/trufi_configuration.dart';
 
-void main() async {
+Future<void> main() async {
```
SamuelRioTz, line 8: why we need async function here?

md-weber: We wait directly for the globalCfg 
grafik (Line 12)

SamuelRioTz: ahh ok, I think we need to handle these async function possible errors, because if something goes wrong, the app will just show black screen.
for now I think is enough just add some FIXME comment

md-weber: Yeah, I guess in the long run I would love to get rid of the app_config file and more create a real class that can be overwritten, then we can add some code comments what which settings is etc. Also, it would be a bit more Typesafe.
I added a `TODO`

SamuelRioTz (approved):

md-weber:
Do not merge yet, I found a bug.

lib/pages/home.dart
```diff
@@ -577,8 +577,8 @@ class HomePageStateData {
     return {
       _fromPlace: fromPlace?.toJson(),
       _toPlace: toPlace?.toJson(),
-      _plan: plan.toJson(),
-      _ad: ad.toJson(),
+      _plan: plan?.toJson(),
+      _ad: ad?.toJson(),
```
md-weber, line 581: I forgot during the refactoring to add the null safety back in place.
Original Code in Master
```dart
      _FromPlace: fromPlace != null ? fromPlace.toJson() : null,
      _ToPlace: toPlace != null ? toPlace.toJson() : null,
      _Plan: plan != null ? plan.toJson() : null,
      _Ad: ad != null ? ad.toJson() : null,
```

SamuelRioTz (approved):

PR #420, opened 2021-04-19
🛠 [Refactoring] State Management improvement with the Bloc Package

md-weber:
The current state management solution is close to the Bloc Package with a lot of self implementation.

The Cubit Library allows a lot of easy state management. Additionally, I tried to set up some more separation of concerns with layers.

- Bloc => Logic and UI State
- Repository => Where the Data Comes from
- UI => Widgets Flutter Layer

It would be amazing to get some feedback before I go further with that. Also it is way better testable :)


lib/location/location_search_delegate.dart
```diff
@@ -323,16 +325,17 @@ class _SuggestionList extends StatelessWidget {
   }
 
   Widget _buildSearchResultList(BuildContext context) {
-    final requestManagerBloc = RequestManagerBloc.of(context);
+    final requestManagerBloc = BlocProvider.of<RequestManagerBloc>(context);
```
md-weber, line 328:
```suggestion
    final requestManagerBloc = context.read<RequestManagerBloc>().state.correlationId;
```

lib/repository/local_repository.dart
```diff
@@ -0,0 +1,27 @@
+abstract class LocalRepository {
```
md-weber, line 1: This will be useful if we think about #422

mdoelker:
Looks good to me. I like that we decouple the persisted state from the actual implementation and the Bloc pattern and model usage seems sensible. Let's continue walking this path!

lib/blocs/home_page_bloc.dart
```diff
@@ -0,0 +1,47 @@
+import 'dart:convert';
+
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:trufi_core/models/map_route_state.dart';
+import 'package:trufi_core/repository/local_repository.dart';
+import 'package:trufi_core/repository/request_manager.dart';
+
+class HomePageBloc extends Cubit<MapRouteState> {
+  LocalRepository sharedPreferencesRepository;
```
mdoelker, line 9: The variable is unnecessarily named after the concrete storage strategy.

md-weber:
I worked now quite some time to fix the issues with the animations finally. There are a lot of problems in that area. The last thing missing is the state inside the PlanPage to avoid jumping if you swap the locations. 

fixed_animations

I want to finalize this PR now (add some tests) for the first couple of changes and create a new one for the other, or do you think it would be better to create one large one?

lib/blocs/home_page_bloc.dart
```diff
@@ -0,0 +1,74 @@
+import 'dart:convert';
+
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:trufi_core/models/map_route_state.dart';
+import 'package:trufi_core/repository/local_repository.dart';
+import 'package:trufi_core/repository/request_manager.dart';
+import 'package:trufi_core/trufi_models.dart';
+
+class HomePageBloc extends Cubit<MapRouteState> {
```
GustavoTCh, line 9: A suggestion, We could rename from HomePageBloc to HomePageCubit, naming it Cubit or Bloc is not relevant? 👍

lib/blocs/preferences_cubit.dart
```diff
@@ -0,0 +1,83 @@
...
+    String correlationId = await localRepository.getCorrelationId();
+
+    // Generate new UUID if missing
+    if (correlationId == null) {
+      correlationId = Uuid().v4();
+      await localRepository.saveCorrelationId(correlationId);
+    }
+  }
+
+  Future<void> _loadLanguageCode() async {
```
GustavoTCh, line 54: This method can have the same structure as the other "_loaders":
```
emit(
      state.copyWith(
        languageCode: await localRepository.getLanguageCode() ?? defaultLanguageCode,
      ),
    );
```
because this class has a static property defaultLanguajeCode.

lib/blocs/preferences_cubit.dart
```diff
@@ -0,0 +1,83 @@
...
+    emit(state.copyWith(loadOnline: loadOnline));
+  }
+
+  Future<void> updateStateHomePage(String stateHomePage) async {
+    if (stateHomePage == null) {
+      await localRepository.deleteStateHomePage();
+    } else {
+      await localRepository.saveStateHomePage(stateHomePage);
+    }
+    emit(state.copyWith(stateHomePage: stateHomePage));
```
GustavoTCh, line 41: when stateHomePage equals null, copyWith receives a null parameter and then the "emit" is called with the same state. Is this correct behavior ?, or do we need to emit a state like "DeleteStateHomePage"?

GustavoTCh (commented):
I gave a quick read to all the code except for the testing part, it looks great. 👍

lib/blocs/preferences_cubit.dart
md-weber: Regarding if it is the `same` state, I guess not. `stateHomePage` comes as a parameter into the function, and if it would be null, the new state will be `null` for the stateHomePage`.

Good question. Regarding the code, I tried to keep it, for the most part, as close to the original as possible. It got deleted in a lot of places by setting a null value. Actually, I guess at the moment I removed the null possibility from the HomePageState. We have to keep an eye on that if that has any bad side effects.

lib/blocs/home_page_bloc.dart
md-weber: Good catch :)

lib/blocs/preferences_cubit.dart
GustavoTCh: Great, I can do "manual tstiong" today 👍 .

GustavoTCh: I did "manual tests" for the main trufi features, I found no critical bugs, only this.
https://github.com/trufi-association/trufi-core/pull/420#issuecomment-824842127

https://user-images.githubusercontent.com/42806689/115804675-79a00080-a3b1-11eb-982f-69b49b79ca8e.mov

GustavoTCh: I did "manual testing" in iOS too, I found no critical bugs, only this.
#420 (comment).

GustavoTCh: I found a bug(IOS and Android) in the "Your Places" page, when I am clicking a place it does not search my route.

https://user-images.githubusercontent.com/42806689/115813154-a576b280-a3c0-11eb-80f8-3983d7b24997.mov

lib/blocs/home_page_cubit.dart
```diff
@@ -0,0 +1,74 @@
...
+  LocalRepository localRepository;
+  RequestManager requestManager;
+
+  HomePageBloc(this.localRepository)
+      : super(MapRouteState(
+          isFetching: false,
+          showSuccessAnimation: false,
+        ));
+
+  Future<void> loadFromSharedPreferences() async {
```
GustavoTCh, line 19: We need to call this method for load this Last-MapRouteState, (when we write the MapRouteState with localRepository this MapRouteState saved)) 

```
  Map<String, dynamic> toJson() {
    return {
      _fromPlace: fromPlace?.toJson(),
      _toPlace: toPlace?.toJson(),
      _plan: plan?.toJson(),
      _ad: ad?.toJson(),
    };
  }
```

 isFetching:null and showAnimation: null, this value do error in PlanPageState

Solution:
1)call method in constructor.

```
  HomePageBloc(this.localRepository)
      : super(MapRouteState(
          isFetching: false,
          showSuccessAnimation: false,
        )) {
    loadFromSharedPreferences();
  }
```

2) We don't called _fetch(method in HomePage) then We define defectValue in MapRouteState.

```
 MapRouteState({
    this.fromPlace,
    this.toPlace,
    this.plan,
    this.ad,
    this.isFetching = false,
    this.showSuccessAnimation = false,
    this.currentFetchAdOperation,
    this.currentFetchPlanOperation,
  }); 
```

or  call loadFromSharedPrefecrences with (CopyWith)
```
  Future<void> loadFromSharedPreferences() async {
    final jsonString = await localRepository.getStateHomePage();

    if (jsonString != null && jsonString.isNotEmpty) {
      emit(
        MapRouteState.fromJson(jsonDecode(jsonString) as Map<String, dynamic>)
        .copyWith(
          isFetching: false,
          showSuccessAnimation: false,
        ),
      );
    }
  }
```

lib/pages/saved_places.dart
```diff
@@ -149,7 +154,7 @@ class SavedPlacesPageState extends State<SavedPlacesPage> {
   }
 
   Future<void> _showCurrentRoute(TrufiLocation toLocation) async {
-    final HomePageStateData dataRoute = HomePageStateData();
+    final MapRouteState dataRoute = MapRouteState();
```
GustavoTCh, line 157: https://github.com/trufi-association/trufi-core/pull/420/files#r618910843 (here error in method _showCurrentRoute)
Before, we updated the "StateHomePage", so every time the HomePage is built "_fetch" is called (in initState method), (it was my mistake, but I needed to add code without modifying the HomePage), now we can add a (`fetch (TrufiLocation to, TrufiLocation from)`) to the HomePageCubit and add this method here.

GustavoTCh (commented):
I found 2 bugs.

1. HomePageBloc don't call the method "loadFromSharedPreferences".
2. SavedPlacesPage, https://github.com/trufi-association/trufi-core/pull/420#discussion_r618910843

lib/blocs/home_page_cubit.dart
md-weber: Fantastic point, thank you for the catch. I changed the toJson method slightly so we have now a possibility to save the null values to false. This way we can assure that there should never be a null state in the DB. I guess that would be the cleanest way.

I also tried to use https://pub.dev/packages/json_serializable the only problem for this class are `currentFetchAdOperation` and `currentFetchPlanOperation`. If we can remove them later we could remove the whole toJson / fromJson and let it handle by the build runner.

lib/pages/saved_places.dart
md-weber: I am not completely sure if I understand correctly. I figured out that the MapRouteState was not immutable. Now I changed that too, and I hope it should already fix the behaviour.

lib/blocs/home_page_cubit.dart
GustavoTCh: it's a great idea to use  https://pub.dev/packages/json_serializable, What do you think of https://pub.dev/packages/hydrated_bloc?

md-weber: That would be even better I guess. 👍 But I would do it in the next PR.

lib/pages/saved_places.dart
GustavoTCh: Excuse me, I did not explain it well.

GustavoTCh: It should work like this(It is very slow in the emulator)

https://user-images.githubusercontent.com/42806689/115844334-bbe73300-a3ed-11eb-8292-fc77da9fd752.mov


but It should work like this(bug)

https://user-images.githubusercontent.com/42806689/115813154-a576b280-a3c0-11eb-80f8-3983d7b24997.mov

lib/blocs/home_page_cubit.dart
GustavoTCh: Good :), jejeje I could add code(feats/issues) after this PR ends 👍

mdoelker:
> I worked now quite some time to fix the issues with the animations finally. There are a lot of problems in that area. The last thing missing is the state inside the PlanPage to avoid jumping if you swap the locations.
> 
> fixed_animations
> 
> I want to finalize this PR now (add some tests) for the first couple of changes and create a new one for the other, or do you think it would be better to create one large one?

I think it's fine to solve this in a follow-up issue.

lib/pages/saved_places.dart
md-weber: This should now work again properly. I needed to take the fetch method into the HomePageCubit. I added a lot of todo´s because the fetch method is a beast. 

If we use the fetch correctly, we can also get rid of the other problem.

SamuelRioTz:
The location error pop-up when you press the location button inside the search bar is with the wrong theme
Simulator Screen Shot - iPhone 12 mini - 2021-04-27 at 02 46 15
Simulator Screen Shot - iPhone 12 mini - 2021-04-27 at 02 46 31

SamuelRioTz:
After getting not route the loader goes into an infinite loop.
I am not sure when this issue was generated, maybe is better if we create issues instead comments.

Simulator Screen Shot - iPhone 12 mini - 2021-04-27 at 02 53 53

simulator_screenshot_97E16DD2-A665-45C9-8D7C-E82CB09B7E08

md-weber:
Yeah, I guess it makes sense if there are so many problems to create issues for each of them and fix them one by one as part of the project.

SamuelRioTz (approved):

PR #427, opened 2021-04-28
[BugFix] Stop fetching after the plan receives an error

### LINKED ISSUE #425: If route is not findable loading spinner goas into infinity loop
**Describe the bug**
After getting the not route the loader goes into an infinite loop.
I am not sure when this issue was generated, maybe is better if we create issues instead of comments.

**To Reproduce**
1. Select a route that is not followable and try another destination
2. Search for a new route
3. Loading spinner is visible for invinity

**Expected behavior**

**Screenshots**

Found in #420

md-weber:
After the error occurs, the fetching stops and the loading spinner is not visible anymore.

Resolves #425 

Currently, this part is not testable until we clean up the fetch function in the HomePageCubit.

### DIFF
--- a/lib/blocs/home_page_cubit.dart
+++ b/lib/blocs/home_page_cubit.dart
@@ -148,6 +148,7 @@ class HomePageCubit extends Cubit<MapRouteState> {
         if (plan == null) {
           throw "Canceled by user";
         } else if (plan.hasError) {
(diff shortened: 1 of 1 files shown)


SamuelRioTz (approved):
Tested and it works as expected

GustavoTCh (commented):
It's working 👍

PR #428, opened 2021-04-28
[BugFix] Removed a lot of context and replaced with the theme

### LINKED ISSUE #424: Wrong Theme of the error pop-up of Location Service
**Describe the bug**
The location error pop-up when you press the location button inside the search bar is with the wrong theme

**To Reproduce**
Steps to reproduce the behaviour:
1. Select Your Location Button on Search Screen

**Expected behaviour**
Should look exactly like the one if you click on the GPS icon.

**Screenshots**

Introduces by #420

md-weber:
The theme needed to be passed down because the search field did not get the correct Theme data from its context. 

Resolves #424

### DIFF
--- a/lib/location/location_search_delegate.dart
+++ b/lib/location/location_search_delegate.dart
@@ -65,7 +65,10 @@ class LocationSearchDelegate extends SearchDelegate<TrufiLocation> {
 
   @override
   Widget buildSuggestions(BuildContext context) {
+    final theme = Theme.of(context);
+
     return _SuggestionList(
+      theme: theme,
       query: query,
       onSelected: (TrufiLocation suggestion) {
         _result = suggestion;
@@ -208,6 +211,7 @@ class _SuggestionList extends StatelessWidget {
     @required this.locationSearchBloc,
     @required this.savedPlacesBloc,
     @required this.appBarTheme,
+    @required this.theme,
   });
 
   final HistoryLocationsBloc historyLocationsBloc;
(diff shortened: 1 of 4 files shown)


GustavoTCh (commented):
It's a good solution.

lib/location/location_search_delegate.dart
```diff
@@ -65,7 +65,10 @@ class LocationSearchDelegate extends SearchDelegate<TrufiLocation> {
 
   @override
   Widget buildSuggestions(BuildContext context) {
+    final theme = Theme.of(context);
+
```
SamuelRioTz, line 69: why the theme is passed as parameter, I thought it should be part of the context? maybe this issues is more related with dialog build context issues

GustavoTCh: https://github.com/trufi-association/trufi-core/blob/73b00220cbcf21cabdd7aff8f9e9855d2adf3a9f/lib/location/location_search_delegate.dart#L31 

The LoactionSearchDelegate defines the theme, background = white.

SamuelRioTz: so I think passing theme as parameter is not the good way, maybe we can override the LocationSearchDelegate background after used

md-weber: Yeah, I am also not happy with the solution. I tried it lower in the widget tree but as soon I get one level down it directly get overwritten from somewhere. The problem is that the SearchDelegate overwrites the Theme of the context and uses its own. 

Currently, I did not have a solution for that and actually did not wanted to invest so much time. But I can investigate where the Problem happens.

GustavoTCh (approved):

md-weber:
I would close this PR in favor for the discussed solution with @SamuelRioTz

PR #430, opened 2021-04-28
[Refactor] Cleanup LocationProviderBloc

md-weber:
The location provider is now in the new bloc system and tested.


lib/blocs/location_provider_cubit.dart
```diff
@@ -0,0 +1,52 @@
...
+    final LocationPermission status = await Geolocator.checkPermission();
+
+    if (status == LocationPermission.always ||
+        status == LocationPermission.whileInUse) {
+      final Position position = await Geolocator.getCurrentPosition(
+          desiredAccuracy: LocationAccuracy.high);
+
+      return position != null
+          ? LatLng(position.latitude, position.longitude)
+          : null;
```
SamuelRioTz, line 47: `getCurrentPosition()` will never return a null value, instead it will generate Throws:

> Throws a [TimeoutException] when no location is received within the supplied [timeLimit] duration. Throws a [PermissionDeniedException] when trying to request the device's location when the user denied access. Throws a [LocationServiceDisabledException] when the user allowed access, but the location services of the device are disabled.

for now I think is ok if we just add `.catchError((error) => null)` to ignore throws.

SamuelRioTz: The throws handling should be handle at the top level, but the error pop ups appears when the location is null, maybe it could part of the error- pop ups improvements

md-weber: I would not recommend swallowing throws. https://en.wikipedia.org/wiki/Error_hiding

SamuelRioTz: the problem is if `getCurrentPosition()` throws an error instead returning null, the error pop up will never appears

lib/location/location_search_delegate.dart
```diff
@@ -639,18 +639,20 @@ class _SuggestionList extends StatelessWidget {
...
-        location: location,
-        addToHistory: false,
+    final location =
+        await context.read<LocationProviderCubit>().getCurrentLocation();
+
+    if (location == null) {
+      showDialog(
+        context: context,
+        builder: (context) => buildAlertLocationServicesDenied(context),
       );
```
SamuelRioTz, line 649: since the location will never be null, I think instead we should handle the throw error.

SamuelRioTz (approved):

lib/location/location_search_delegate.dart
SamuelRioTz: @md-weber  fixed here [PR 434](https://github.com/trufi-association/trufi-core/pull/434)

lib/blocs/location_provider_cubit.dart
md-weber: Great catch :) thank you!

SamuelRioTz: yes

SamuelRioTz: #435

SamuelRioTz (approved):

GustavoTCh (approved):

PR #431, opened 2021-04-28
[Feat] Add support to graphql with package

GustavoTCh:
It is a implementation with GraphQL  `graphql: ^4.0.1`  package.
The package does not cause problems with the other current dependencies

If a future we need to work with (Cached, mutations), It is better to work with the package.


md-weber:
Should we even proceed with the GraphQL solution with the package? I thought we decided on the basic HTTP way?

SamuelRioTz:
@GustavoTCh  I think this PR should just a draft

PR #432, opened 2021-04-28
[Feat] Add support to graphql with simple HTTP request

GustavoTCh:
GraphQL queries based on http requests. Without the use of any package.

You need change from trufiCore-Server to standtnavi-DevServer in the app-config.json file

Advantage:

Simple request.
We do not depend on any package
Disadvantages:

if we need to do some kind of (CACHE) we need to implement it manually.
This getPlan work. but the data response of stadNaviaServer and trufiServer is a little different.
We need to define the other types of queries(WALK, CAR, RAIL...)


example/lib/main.dart
```diff
@@ -141,6 +141,7 @@ Future<void> main() async {
   trufiCfg.url.twitter = globalCfg.getValue<String>("urlTwitter");
   trufiCfg.url.instagram = globalCfg.getValue<String>("urlInstagram");
   trufiCfg.url.share = globalCfg.getValue<String>("urlShare");
+  trufiCfg.url.isOtpGraphQL = globalCfg.getValue<bool>("isOtpGraphQL");
```
SamuelRioTz, line 144: instead a bool, I think is better if we just use a string with the type. e.g. `otpType = "GraphQl" | "default"`

md-weber: Or maybe create an enum? - Should we even use the GlobalCfg or the TrufiConfiguration signleton?

SamuelRioTz: even better, I was think it will come from the `config.json` file, but this is not the case.

lib/repository/online_graphql_repository/online_graphql_repository.dart
```diff
@@ -0,0 +1,125 @@
...
+      "query": '''
+        query {
+          $queryPlan
+        }
+      '''
+    };
+
+    final response = await _fetchRequest(request, body);
+    if (response.statusCode == 200) {
+      return compute(_parsePlan, utf8.decode(response.bodyBytes));
```
md-weber, line 98: I would try to avoid compute functions and only use it if we really parse very very large jsons.

example/lib/main.dart
md-weber: I guess at the moment it comes from the app_config.json file but we could add it to the TrufiConfiguration. 
https://github.com/trufi-association/trufi-core/blob/master/lib/trufi_configuration.dart#L31

lib/trufi_configuration.dart
```diff
@@ -3,6 +3,8 @@ import 'package:flutter/material.dart';
 import 'package:latlong/latlong.dart';
 import 'package:trufi_core/l10n/trufi_custom_localization.dart';
 
+enum TypeServer { defaultServer, graphQLServer }
```
md-weber, line 6: Nit: I would call the enum ServerType => The type of the server.

lib/trufi_app.dart
```diff
@@ -98,7 +99,9 @@ class TrufiApp extends StatelessWidget {
         BlocProvider<RequestManagerCubit>(
           create: (context) => RequestManagerCubit(
             OfflineRepository(),
-            OnlineRepository(),
+            (TrufiConfiguration().generalConfiguration.typeServer == TypeServer.defaultServer)
```
md-weber: Do we need the braces around that statement?
```suggestion
            TrufiConfiguration().generalConfiguration.typeServer == TypeServer.defaultServer
```

lib/repository/online_graphql_repository/queries.dart
```diff
@@ -0,0 +1,141 @@
+import 'package:meta/meta.dart';
+
+String getPlanComplete({
```
md-weber, line 3: Could have a tiny test that executes and gets a String to avoid typos

lib/models/utils_models.dart
```diff
@@ -0,0 +1,10 @@
+class EnumValues<T> {
```
md-weber, line 1: I believe I know what you want to do with that (Strings in an Enum) but rather that I would recommend to write an extension method: https://stackoverflow.com/a/58552304/3690032

lib/repository/online_graphql_repository/online_graphql_repository.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  @override
+  CancelableOperation<Ad> fetchAd(BuildContext context, TrufiLocation to) {
+    return _fetchCancelableAd(context, to);
+  }
+
+  @override
+  CancelableOperation<Plan> fetchCarPlan(
+      BuildContext context, TrufiLocation from, TrufiLocation to) {
+    return _fetchCancelablePlan(from, to, "TRANSIT,WALK");
```
md-weber, line 34: Would it make sense to add also here the enums? Transit, Walk etc.?

md-weber (commented):
It is hard for me to grasp the GraphQL part, but I trust you there that it works, maybe we could one or two more tests 👍

lib/models/utils_models.dart
GustavoTCh: Thanks, @md-weber I will review the documentation :)

lib/repository/online_graphql_repository/plan_graphql_model.dart
```diff
@@ -1,12 +1,41 @@
 import 'package:flutter/material.dart';
-import 'package:trufi_core/models/utils_models.dart';
 import 'package:trufi_core/trufi_models.dart';
 
-enum _Mode { walk, rail, bus }
+enum TransportMode { walk, rail, bus, transit, car }
+
+TransportMode _getEnumMode(String enumS) {
+  switch (enumS) {
```
md-weber, line 7: Instead of a switch, I like the idea of a map, what do you think?
https://dartpad.dartlang.org/c4001d907d6a420cafb2bc2c2507f72c

GustavoTCh:
I need to implement the unit Test for GraphQL integration, it's working with Car-WALK and TRANSIT(Transport Public for example BUS-RAIL) - WALK how Trufi-App.

Result with large travel(the route use Light Rail)

Screenshot_1619708807

Screenshot_1619708800

lib/repository/online_graphql_repository/plan_graphql_model.dart
GustavoTCh: It's good to use a map in the extensionEnum class if we do _getEnumMode a map, We need to control the response null when we need to use the map we need to put "??" for control value null.

can the map help me to control default values how the switch?

GustavoTCh: enum TransportMode { walk, rail, bus, transit, car }

const transportModeMap= {
  'WALK': TransportMode.walk,
  'BUS': TransportMode.bus,
  'CAR': TransportMode.car,
  'LIGHT RAIL': TransportMode.rail,
  'TRANSIT': TransportMode.transit,
};

TransportMode _getTransportMode(String enumText) 
  => transportModeMap[enumText] ?? TransportMode.walk;

extension TransportModeExtension on TransportMode {
  static const names = {
    TransportMode.walk: 'WALK',
    TransportMode.bus: 'BUS',
    TransportMode.car: 'CAR',
    TransportMode.rail: 'LIGHT RAIL',
    TransportMode.transit: 'TRANSIT',
  };
  String get name => names[this] ?? '';
}

lib/repository/online_graphql_repository/plan_graphql_model.dart
```diff
@@ -0,0 +1,217 @@
...
+          .map(
+            (itinerary) => itinerary.toPlanItinerary(),
+          )
+          .toList(),
+      error: itineraries.isEmpty ? PlanError(404, 'Not found routes') : null,
+    );
+  }
+}
+
+class _Location {
```
md-weber, line 69: What is the difference to the TrufiLocation?

md-weber, line 69: Can we make these parts reusable, I guess a Location will never be that unique and should be used at multiple parts, maybe we can even upgrade the TrufiLocation with a Location object.

SamuelRioTz (approved):

SamuelRioTz (approved):

test/repository/online_graphql_repository/queries_test.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:trufi_core/repository/online_graphql_repository/queries.dart'
+    as queries;
+
+void main() {
+  group("Queries", () {
+    test("should return the query with the given parameters", () {
```
md-weber, line 7: Why did we only test one query?

lib/repository/online_graphql_repository/queries.dart
```diff
@@ -0,0 +1,149 @@
...
+            elevation
+          }
+        },
+      },
+    }
+  }
+''';
+}
+
+String getCustomPlan({
```
md-weber, line 90: There are two :)

PR #433, opened 2021-04-28
Fix/android location permissions

SamuelRioTz:


md-weber (approved):
Great idea to do that in this PR 👍

PR #438, opened 2021-04-29
Improvement/fetch manager

SamuelRioTz:
should be merged after #437 
improvement and clean code for the` fetch plan function`

PR #439, opened 2021-04-29
fix show emptyMap

GustavoTCh:
We need to show the EmptyMap, but the cubit state does not allow us to have the plan in null. 

Edit: The first time "Plan" is null, but once it is given a value, it can no longer be null


GustavoTCh:
Fixed in https://github.com/trufi-association/trufi-core/pull/438

PR #442, opened 2021-04-30
BugFix-Fetch Plan by car

SamuelRioTz:
After fail the first request and request it again(routes with cars), the `TrufiLocalization.of(context)` does not contains any localization, due wrong context was sent on the last request.


md-weber (commented):
Good catch 👍

md-weber (approved):

PR #452, opened 2021-05-06
Feat/Settings Panel

GustavoTCh:
EDIT: My fault, the commits to review are those of May 6, the other commits are already in the Master branch.

Screen Shot 2021-05-06 at 07 20 15

- Created a fetchAdvancedPlan(), for us to send parameters in the query.
- "SettingPanel"  was created(UI basic and complete behavior how in DigitTransitUI)
- SettingPanel only works with graphql endpoint but does not break queries via traditional OTP.

- The button setting is visible when there is a route painted on the map 

Screenshot_1620295672

- If something is modified in the configuration panel, then when going back a new query will be made with the new parameters.

Screenshot_1620295674


SamuelRioTz:
@GustavoTCh I think we need to move the settings buttons, because the settings is related with the query instead with the map
Screen Shot 2021-05-06 at 06 54 09

GustavoTCh:
@SamuelRioTz  I think the same as you, but for the moment I copied the behavior of (StadtnaviUINext), only when we have a route on the map or a route empty can the SettingPanel be accessed. I think this can be modified when we have the UI / UX

Screen Shot 2021-05-06 at 07 16 35

SamuelRioTz:
@GustavoTCh but on the example the settings button is part of the search bar, is not part of the map

lib/services/plan_request/online_graphql_repository/queries.dart
```diff
@@ -151,7 +149,126 @@ String getCustomPlan({
...
 
-String _parseTransportModes(List<TransportMode> transportModes) {
-  final modes = transportModes.map((e) => '{mode:${e.name}}').join(',');
-  return '[$modes]';
+String getPlanAdvanced({
+  @required double fromLat,
+  @required double fromLon,
+  @required double toLat,
+  @required double toLon,
+  // Research about use
```
SamuelRioTz, line 157: is it a TODO? if yes, pls add a TODO: tag

lib/plan/setting_panel/setting_panel.dart
```diff
@@ -0,0 +1,284 @@
...
+          builder: (blocContext, state) {
+            return ListView(
+              children: <Widget>[
+                const SizedBox(
+                  height: 10,
+                ),
+                CustomExpansionTile(
+                  title: "Walking speed",
+                  options: const [
+                    'Slow',
```
md-weber, line 41: Looks like a nice place for an enum again :)

lib/plan/setting_panel/setting_panel.dart
```diff
@@ -0,0 +1,284 @@
...
+                    'Calm',
+                    'Average',
+                    'Prompt',
+                    'Fast',
+                  ],
+                  textSelected: _getTextWalkingSelect(context,state.typeWalkingSpeed),
+                  onChanged: (value) {
```
md-weber, line 48: If the option is the enum you can just call the settingPanelCubit.setWalkingSpeed(value)

lib/plan/setting_panel/setting_panel.dart
```diff
@@ -0,0 +1,284 @@
...
+                      default:
+                    }
+                  },
+                ),
+                const Divider(thickness: 2),
+                CustomSwitchTile(
+                  title: 'Avoid walking',
+                  value: state.avoidWalking,
+                  onChanged: settingPanelCubit.setAvoidWalking,
+                ),
```
md-weber, line 74: Not sure but this could be a list of objects and then we use the ListView.seperated to create the different once.

lib/plan/setting_panel/setting_panel.dart
```diff
@@ -0,0 +1,284 @@
...
+                const SizedBox(height: 10),
+              ],
+            );
+          },
+        ),
+      ),
+    );
+  }
+
+  String _getTextWalkingSelect(BuildContext context, WalkingSpeed selected) {
```
md-weber, line 263: I would do that with an extension method on the enum.

lib/plan/setting_panel/setting_panel_cubit.dart
```diff
@@ -0,0 +1,94 @@
...
+    }
+    await updateMapRouteState(state.copyWith(bikeRentalNetworks: newList));
+  }
+
+  Future<void> setWalkingSpeed(WalkingSpeed walkingSpeed) async {
+    await updateMapRouteState(state.copyWith(typeWalkingSpeed: walkingSpeed));
+  }
+
+  // ignore: avoid_positional_boolean_parameters
+  Future<void> setAvoidWalking(bool avoidWalking) async {
```
md-weber, line 62: Instead of ignoring it you could just do it.
```suggestion
  // ignore: avoid_positional_boolean_parameters
  Future<void> setAvoidWalking({bool avoidWalking}) async {
```

lib/plan/setting_panel/setting_panel_state.dart
```diff
@@ -0,0 +1,228 @@
+part of 'setting_panel_cubit.dart';
+
+enum WalkingSpeed { slow, calm, average, prompt, fast }
+enum BikingSpeed { slow, calm, average, prompt, fast }
+enum WalkBoardCost { defaultCost, walkBoardCostHigh }
+enum BikeRentalNetwork { taxi, carSharing, regioRad }
+enum OptimizeType { quick, safe, flat, greenWays, triangle, transfers }
```
md-weber, line 7: I would not create all the enums inside of the state, it would be better to move them to a enum file or create for each enum a separated file.

lib/repository/local_repository.dart
```diff
@@ -24,4 +24,9 @@ abstract class LocalRepository {
   Future<void> deleteStateHomePage();
 
   Future<String> getStateHomePage();
+
+  // Operations about StateSettingPanel 
+  Future<void> saveStateSettingPanel(String stateHomePage);
```
md-weber, line 29: The attribute here is more a stateSettingJson right?
```suggestion
  Future<void> saveStateSettingPanel(String stateSettingJson);
```

lib/repository/shared_preferences_repository.dart
```diff
@@ -94,4 +95,16 @@ class SharedPreferencesRepository implements LocalRepository {
     final preference = await _sharedPreferences;
     return preference.getString(_stateHomePageKey);
   }
+
+  @override
+  Future<String> getStateSettingPanel()async {
```
md-weber, line 100: Nit
```suggestion
  Future<String> getStateSettingPanel() async {
```

lib/services/plan_request/online_graphql_repository/queries.dart
```diff
@@ -151,7 +149,126 @@ String getCustomPlan({
 ''';
 }
 
-String _parseTransportModes(List<TransportMode> transportModes) {
-  final modes = transportModes.map((e) => '{mode:${e.name}}').join(',');
-  return '[$modes]';
+String getPlanAdvanced({
```
md-weber, line 152: Test?

lib/widgets/custom_switch_tile.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:flutter/material.dart';
+
+class CustomSwitchTile extends StatelessWidget {
```
md-weber, line 3: Looks a bit small for an own widget. Is it essential?

lib/widgets/map_setting_button.dart
```diff
@@ -0,0 +1,41 @@
...
+class MapSettingButton extends StatelessWidget {
+  const MapSettingButton({Key key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return FloatingActionButton(
+      mini: true,
+      backgroundColor: Theme.of(context).backgroundColor,
+      onPressed: () async {
+        final oldSettingPanelCubit = context.read<SettingPanelCubit>().state.copyWith();
```
md-weber, line 20: Why do you call here `copyWith`?

md-weber:
> 
> 
> @SamuelRioTz I think the same as you, but for the moment I copied the behaviour of (StadtnaviUINext), only when we have a route on the map or a route empty can the SettingPanel be accessed. I think this can be modified when we have the UI / UX
> 
> Screen Shot 2021-05-06 at 07 16 35

Sounds like something we should discuss with our Design Responsible :) https://trufiapp.slack.com/team/U01384X6971

lib/plan/setting_panel/setting_panel.dart
GustavoTCh: Thanks :) , it is changed

lib/plan/setting_panel/setting_panel.dart
GustavoTCh: Thanks :) , it is changed

lib/plan/setting_panel/setting_panel.dart
GustavoTCh: There will also be changes to the user interface, it would be good to wait.

lib/plan/setting_panel/setting_panel.dart
GustavoTCh: Thanks :) , it is changed

lib/plan/setting_panel/setting_panel_cubit.dart
GustavoTCh: Thanks :) , it is changed

lib/plan/setting_panel/setting_panel_state.dart
GustavoTCh: Thanks :) , it is changed

lib/services/plan_request/online_graphql_repository/queries.dart
GustavoTCh: Thanks :) , it is changed

lib/widgets/custom_switch_tile.dart
GustavoTCh: I created it to centralize the UI(how theme and TextStyle) of all SwitchTile.

lib/widgets/map_setting_button.dart
GustavoTCh: Before it was necessary but now not, I forgot to remove it, thanks :)

SamuelRioTz:
Screen Shot 2021-05-07 at 08 23 50
@GustavoTCh

SamuelRioTz (approved):

PR #453, opened 2021-05-07
support url feedback

GustavoTCh:
If you don't define "emailFeedback" in config.json, the application would use "urlRouteFeedback".

- If the emailFeedback exists the app used the email icon,

Screen Shot 2021-05-07 at 07 30 28

- If the emailFeedback does not exist, the app use the feedback icon,

Screen Shot 2021-05-07 at 07 29 04


lib/pages/feedback.dart
```diff
@@ -1,113 +1,88 @@
...
+            url = feedBack.body;
+          }
+          if (await canLaunch(url)) {
+            await launch(url);
+          } else {
+            await showDialog(
+              context: context,
+              builder: (BuildContext context) {
+                return AlertDialog(
+                  // TODO translate
```
md-weber, line 62: To remove this TODO, I created a PR https://github.com/trufi-association/trufi-core/pull/454

lib/trufi_configuration.dart
```diff
@@ -107,3 +108,22 @@ class TrufiConfigurationUrl {
...
+enum FeedBackType { email, url }
+
+class DefinitionFeedBack {
+  /// You can selected the type feedback with enum
+  /// [type] parameter support {email and url} 
+  final FeedBackType type;
+  /// [body] is you url or email for feedback.
+  final String body;
+
+  DefinitionFeedBack({
```
md-weber, line 125: In order to enforce usage rather than just make it "required" you could just use the not named constructor.
```suggestion
   DefinitionFeedBack(this.type, this.body);
```

lib/widgets/trufi_drawer.dart
```diff
@@ -103,11 +103,12 @@ class TrufiDrawerState extends State<TrufiDrawer> {
             localization.menuYourPlaces,
             SavedPlacesPage.route,
           ),
+          if (cfg.configurationDrawer.definitionFeedBack != null) 
           _buildListItem(
             Icons.feedback,
             localization.menuFeedback,
             FeedbackPage.route,
-          ),
+          ) else Container(),
```
md-weber, line 111: Is this container only to show nothing? If so you could just remove the else part:
```dart
if (cfg.configurationDrawer.definitionFeedBack != null)
            _buildListItem(
              Icons.feedback,
              localization.menuFeedback,
              FeedbackPage.route,
            ),
```

lib/trufi_configuration.dart
GustavoTCh: I add the change in a new commit, because we need to remove the "import 'package:meta/meta.dart';"

SamuelRioTz (approved):

PR #455, opened 2021-05-08
Feat/enhance favorites functionality

GustavoTCh:
Trabajé en los siguientes puntos:
 
- The SavedPlaceHome has saved places and favorites places.
- The places has 4  options(change icon, change name, change position and remove).

- Favorite places only have enable the options of (change icon and remove)
Screen Shot 2021-05-08 at 18 13 53

- Implemented the default locations( home and work), these locations are very easy add translates.

- Default places only have  enable the options of (change icon and remove)
Screen Shot 2021-05-08 at 18 14 11

- Enhance the call fetchLocation.
- Fix the animation when we call fetch.

https://user-images.githubusercontent.com/42806689/117554862-40f65d00-b028-11eb-852a-90898f413a1e.mov

- The change position.

https://user-images.githubusercontent.com/42806689/117555032-b4e53500-b029-11eb-98f5-052c3d43302e.mov


lib/pages/saved_places/dialog_select_icon.dart
```diff
@@ -0,0 +1,42 @@
...
+        SizedBox(
+          width: 200,
+          height: 200,
+          child: GridView.builder(
+            itemCount: icons.length,
+            gridDelegate: const SliverGridDelegateWithFixedCrossAxisCount(crossAxisCount: 4),
+            itemBuilder: (BuildContext builderContext, int index) {
+              return InkWell(
+                onTap: () {
+                  Navigator.of(context).pop(icons.keys.toList()[index]);
```
md-weber, line 32: Instead of transferring into a list, you could directly call the elementAt function. `icons.keys.elementAt(index)`

lib/pages/saved_places/dialog_select_icon.dart
```diff
@@ -0,0 +1,42 @@
...
+                },
+                child: Icon(icons.values.toList()[index]),
```
md-weber, line 34:
```suggestion
                child: Icon(icons.values.elementAt(index)),
```

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+        length: 2,
+        child: Column(
+          children: [
+            const SizedBox(height: 10),
+            TabBar(
+              tabs: [
+                Container(
+                  padding: const EdgeInsets.all(10),
+                  child: Text(
+                    // TODO translate
```
md-weber, line 36: I would recommend directly use the localization else we will have a lot of work to find all the locations and understand how we can update them.

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+    final searchLocationsCubit = context.read<SearchLocationsCubit>();
+    final localization = TrufiLocalization.of(context);
+    final theme = Theme.of(context);
+    return Scaffold(
+      appBar: AppBar(title: Text(localization.menuYourPlaces)),
+      body: DefaultTabController(
+        length: 2,
+        child: Column(
+          children: [
+            const SizedBox(height: 10),
```
md-weber, line 30: I would not recommend this, if you press and hold the Tab the Material Ripple effect looks odd. Instead, increase the padding of the children on top by ten.

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+            TabBar(
+              tabs: [
+                Container(
+                  padding: const EdgeInsets.all(10),
```
md-weber, line 34:
```suggestion
                  padding: const EdgeInsets.fromLTRB(10, 20, 10, 10),
```

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+                  child: Text(
+                    // TODO translate
+                    "Saved places",
+                    style: theme.textTheme.bodyText1,
+                  ),
+                ),
+                Container(
+                  padding: const EdgeInsets.all(10),
```
md-weber, line 42:
```suggestion
                  padding: const EdgeInsets.fromLTRB(10, 20, 10, 10),
```

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+                    maxLines: 2,
+                  ),
+                )
+              ],
+            ),
+            const SizedBox(height: 5),
+            Expanded(
+              child: TabBarView(
+                children: [
+                  Scaffold(
```
md-weber, line 56: Flutter Documentation
```
/// Although there are some use cases, like a presentation app that
/// shows embedded flutter content, where nested scaffolds are
/// appropriate, it's best to avoid nesting scaffolds.
```

In our case, we need a FloatingActionButton if we are on the first screen is it really mandatory to create a second `Scaffold`?

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+                    body: BlocBuilder<SearchLocationsCubit, SearchLocationsState>(
+                      builder: (context, state) {
+                        return ListView(
+                          padding: const EdgeInsets.symmetric(horizontal: 16.0),
+                          children: [
+                            const SizedBox(height: 10),
+                            Builder(builder: (_) {
```
md-weber, line 63: Why do we need a Builder here if we do not use the context that it provides?

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+                          ],
+                        );
+                      },
+                    ),
+                    floatingActionButton: FloatingActionButton(
+                      onPressed: () {
+                        _addNewPlace(context);
+                      },
+                      backgroundColor: Theme.of(context).primaryColor,
+                      heroTag: null,
```
md-weber, line 105: Why do we set the heroTag always to null?

lib/pages/saved_places/saved_places.dart
```diff
@@ -0,0 +1,155 @@
...
+    final searchLocationsCubit = context.read<SearchLocationsCubit>();
+    final LatLng mapLocation = await ChooseLocationPage.selectPosition(context);
+    if (mapLocation != null) {
+      final String description = await showDialog(
+        context: context,
+        builder: (BuildContext context) {
+          return const DialogEditText();
+        },
+      );
+      if (description != null) {
```
md-weber, line 145: This case can never happen because it seems that the save button prevents a null value.

md-weber (commented):
Some tests would be good, additionally, I think it would be great if the "EditTextDialog" gets more options. Currently, there is no requirement for it but to change the name and afterwards change the icon feels not really user friendly. Maybe talk about the UI UX with Sandra - @Denis306 .

lib/pages/saved_places/saved_places.dart
GustavoTCh: I used the second scaffold for the fast implementation because I think there will be a change in the UI, I understand your suggestion, and I will use a stack. 👍

lib/pages/saved_places/saved_places.dart
GustavoTCh: Sorry, it's my mistake, I'll pay more attention to these kinds of mistakes

lib/pages/saved_places/saved_places.dart
GustavoTCh: When we use the cancel button or close the DialogEditText, the description is null.

lib/pages/saved_places/saved_places.dart
GustavoTCh: I talked with @SamuelRioTz  and  We think that these translations can change and even some will no longer be used, to avoid the double work of modifying or deleting translations we plan to do them almost at the end.

md-weber: Ok, if you decided as a team I am fine with it.

SamuelRioTz: @md-weber  the translation is also part of the UI/UX changes probably the translations will be changed, so we talked for now is ok skip some translations, especially the new ones

SamuelRioTz (approved):

PR #462, opened 2021-05-12
add support online search

GustavoTCh:
- now We can send an instance of SearchLocationManager how parameter to TrufiApp


lib/location/location_search_delegate.dart
```diff
@@ -202,21 +201,21 @@ class _SuggestionList extends StatelessWidget {
...
+              onStreetTapped,
             ),
+          // if (query.isEmpty)
+          //   _BuildFutureBuilder(
+          //     title: localization.searchTitlePlaces,
+          //     future: locationSearchBloc.fetchPlaces(context),
+          //     iconData: Icons.place,
+          //     onSelected: onSelected,
+          //     onStreetTapped: onStreetTapped,
+          //   ),
```
SamuelRioTz, line 218: we need to avoid keep commented code

lib/trufi_app.dart
```diff
@@ -106,6 +108,8 @@ class TrufiApp extends StatelessWidget {
   /// List of Map Tile Provider
   /// if the list is [null] or [Empty], [Trufi Core] then will be used [OSMDefaultMapTile]
   final List<MapTileProvider> mapTileProviders;
+
+  final SearchLocationManager searchLocationManager;
```
SamuelRioTz, line 112: description is needed here

SamuelRioTz: the default value should included here

SamuelRioTz (approved):

PR #465, opened 2021-05-13
Feat/choose origin destination directly  map

GustavoTCh:
- Fixed the chooseSelectMap
- Implemented choose origin and destination with the long press in the map
- Implemented tests

Resolves https://github.com/stadtnavi/stadtnavi_app/issues/55

### DIFF
--- a/lib/blocs/home_page_cubit.dart
+++ b/lib/blocs/home_page_cubit.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 
 import 'package:async/async.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:latlong/latlong.dart';
 import 'package:trufi_core/entities/ad_entity/ad_entity.dart';
 import 'package:trufi_core/entities/plan_entity/plan_entity.dart';
 import 'package:trufi_core/models/map_route_state.dart';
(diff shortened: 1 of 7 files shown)


lib/blocs/home_page_cubit.dart
```diff
@@ -50,6 +51,16 @@ class HomePageCubit extends Cubit<MapRouteState> {
     emit(newState);
   }
 
+  Future<void> setTappingPlace(LatLng latLng) async {
```
md-weber, line 54: Just an unverified idea.
```dart
Future<void> setTappingPlace(LatLng latLng) async {
  await updateMapRouteState(
    state.copyWith(
        fromPlace: state.fromPlace ?? TrufiLocation.fromLatLng("Map Marker", latLng),
        toPlace: state.toPlace ?? TrufiLocation.fromLatLng("Map Marker", latLng)
    ),
  );
}
```

md-weber (approved):
<3 Tests

lib/blocs/home_page_cubit.dart
GustavoTCh: when state.fromPlace == state.toPlace== null, the setTappingPlace defines the from an to with the same value.

PR #466, opened 2021-05-14
Feat/adding double tap zoom

GustavoTCh:
adding double tap zoom

Resolves https://github.com/stadtnavi/stadtnavi_app/issues/56


md-weber (approved):
I guess this builds on top of #465

PR #474, opened 2021-05-23
❌ [Removed] Removed the weather service of FMI

md-weber:
Removed the Weather Service with FMI, I decided to create a package for that in my free time and will depend on it so we do not have to include all the images etc. for the FMI service. 

Additionally, there is currently a bug in Flutter 2.0.6 and the HTTP service is not able to receive the response. I tested it in Flutter 2.2 and there are works. Therefore I decided, for now, to take it outside of the Trufi Core and will work with [OpenWeatherMap](https://openweathermap.org/price)


md-weber:
The issue even persist with OpenWeatherMap 😞 
```
[VERBOSE-2:ui_dart_state.cc(186)] Unhandled Exception: SocketException: Connection failed, address = api.openweathermap.org, port = 443
```

Currently we are blocked with that issue until Version 2.2 of Flutter.

SamuelRioTz (approved):

PR #521, opened 2021-06-17
Implement car-carpark

GustavoTCh:
- Car and CarPark
Simulator Screen Shot - iPhone 11 Pro Max - 2021-06-18 at 08 50 22


Denis306:
@GustavoTCh If you have the Language in English (Like in the itinerary) the Titel is not being translated: 
image

SamuelRioTz (approved):

PR #633, opened 2022-09-18
[wip] Add and generate Amharic Translations

amenk:
* Extend generation script
* Extend Language list
* Add partial Amharic translation
* Generate files

### DIFF
--- a/lib/base/blocs/localization/trufi_localization_cubit.dart
+++ b/lib/base/blocs/localization/trufi_localization_cubit.dart
@@ -20,6 +20,8 @@ class TrufiLocalizationCubit extends Cubit<TrufiLocalization> {
         return 'Italiano';
       case 'pt':
         return 'Portuguese';
+      case 'am':
+        return 'አማርኛ';
       default:
         return 'English';
     }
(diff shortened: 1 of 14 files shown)


amenk:
@alazar-tekle can you complete the Amharic translation? 
*Important* Only work on the *.arb files in the translation/* folders - do not update the .dart files which are also contained in this branch, they are generated and I can do that.

SamuelRioTz:
@amenk please merge main branch change on you PR, and convert into draf if the PR is not yet ready to merge

amenk:
@alazar-tekle Thank you, looks technically good -- about page is still missing.

Wondering if the About-Trufi should be extended by an About-Addismap section?

@SamuelRioTz  I think this would need local translation overwrites?
Or should we add an "about local additional" block into the core which can be overwritten?

GustavoTCh:
@amenk Did you require a custom About-Page, in Trufi-App exist an example about Custom-Pages, but this is a [Trufi-Custom-Feedback-Page](https://github.com/trufi-association/trufi-app/blob/migration/lib/feedback.dart), for this you require [custom-default-values](https://github.com/trufi-association/trufi-app/blob/migration/lib/default_values.dart), and you need to use the custom-default-values in main file.

PR #782, opened 2025-12-02
Feat/implement trufi core maps

SamuelRioTz:
Summary
- Add new `trufi_core_maps` package with support for flutter_map and maplibre_gl
- Implement 3-layer architecture: Presentation (map widgets), Domain (controllers, layers, entities), and Data (tile system, spatial indexing)
- Include example app demonstrating package features

should be merged after https://github.com/trufi-association/trufi-core/pull/783


SamuelRioTz:
@md-weber 
I was struggling trying to extract some modules from the old trufi_core as new packages, but it's quite difficult. 
I think is better to first focus first on the critical and fundation packages, later I will work on integrate into trufi_core and replace the old maps manager and renderer

PR #864, opened 2026-03-23
fix: filter redundant same-line transfers and add walk routes for short trips

### LINKED ISSUE #859: [Bug] [v5-Cochabamba] Route planner suggests two Line 209 minibuses for a short-distance trip
### 📝 Description

A routing inconsistency was detected in the Trufi route planner. When selecting two locations that are geographically very close to each other, the system suggests a route that requires taking **two different minibuses (Line 209 twice)** with an unnecessary transfer.

Based on the map distance, the suggested route appears inefficient and overly complex for such a short trip. The planner should ideally suggest a **simpler route (single line, walking, or a more direct path)** instead of forcing a transfer between vehicles.

* **Priority:** Medium (Should be improved to optimize routing suggestions).
* **Criticality:** Medium (Functionality works but produces inefficient routes).

### 📱 Test Environment

* **Device:** Desktop PC
* **OS:** Windows 11
* **Browser:** Chrome v145
* **App Version:** Trufi Planner Web (v5 – Cochabamba)

### 🛠 Steps to Reproduce

1. Open the web planner: `https://planner.trufi.app/`
2. Select an origin location near **Laguna de Coña Coña**.
3. Select a destination location slightly east along **Avenida Sexta / Antonio Montes**.
4. Allow the planner to calculate routes.
5. Observe the suggested first route.

### Evidence

Image

### ❌ Actual Behavior

The planner suggests a route that requires:

* Walking to a stop
* Taking **Minibus Line 209**
* Transferring to **another Minibus Line 209**
* Arriving at the destination

This results in an unnecessarily complex trip for a relatively short distance (~7 km shown in the planner with minimal spatial separation on the map).

### ✅ Expected Behavior

The route planner should suggest a **more efficient and realistic option**, such as:

* A **single minibus line without transfer**, or
* A **direct route with fewer segments**, or
* A **short walking route if applicable**.

The system should avoid suggesting **redundant transfers between the same line** when the origin and destination are close.

guss1215:
## Summary
- Fixes #859 — Filters out itineraries where the same transit line appears consecutively (e.g., Line 209 → Line 209), which are inefficient routing artifacts from OTP
- Rebalances itinerary sorting to prioritize fewer transfers and shorter duration over walking distance
- Fetches real street-level walk-only routes from OTP for short-distance trips (< 2.5 km), so users see walking as a practical alternative with actual street geometry

## Files changed
- `routing_engine_request_plan_service.dart` — Added redundant-transfer filter, walk route fetching via OTP, and improved sorting weight
- `home_screen_trufi_screen.dart` — Added optional `walkRouteEndpoint` parameter
- `pubspec.yaml` — Added `http` dependency for walk route requests

## Test plan
- [ ] Search a short-distance route (~500m-1km) → walk-only option appears first with real street path
- [ ] Search a route that previously suggested same line twice → redundant transfer is filtered out
- [ ] Search a long-distance route (> 2.5 km) → no walk-only option injected
- [ ] Verify transit routes still sort correctly (fewer transfers preferred)
- [ ] Verify mobile routing is unaffected

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/home_screen_trufi_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/home_screen_trufi_screen.dart
@@ -38,11 +38,6 @@ class HomeScreenTrufiScreen extends TrufiScreen {
   /// Provides the route code to allow navigation to route details screen.
   final void Function(BuildContext context, String routeCode)? onRouteTap;
 
-  /// Optional OTP endpoint for fetching walk-only routes with real street
-  /// geometry. When provided, walking alternatives are automatically
-  /// included in route results for short-distance trips.
-  final String? walkRouteEndpoint;
-
   /// Static initialization for the module.
   /// Call this once at app startup before using any HomeScreen functionality.
   static Future<void> init() async {
@@ -55,7 +50,6 @@ class HomeScreenTrufiScreen extends TrufiScreen {
     this.onItineraryDetails,
     this.onStartNavigation,
     this.onRouteTap,
-    this.walkRouteEndpoint,
   }) {
     _repository = repository ?? HomeScreenRepositoryImpl();
   }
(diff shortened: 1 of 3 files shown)


SamuelRioTz:
must be fixed in [planner](https://github.com/trufi-association/trufi-server-planner)

PR #880, opened 2026-04-22
fix to navigation delegate

armandohackcode:
- home_screen_trufi_screen
- navigation_screen
- The necessary libraries define a delegate navigation


SamuelRioTz:
already fixed here https://github.com/trufi-association/trufi-core/pull/877

PR #894, opened 2026-05-08
fix: hide departure-time chip in wide/web layout + Provider<AppConfiguration?>

SamuelRioTz:
## Summary

Two related fixes that together make `routingTimeOverride` work end-to-end on the web / wide-tablet layout.

### 1. Wide-layout chip duplicate (the real bug)

`HomeScreen` has two layouts — mobile and wide/web. The 5.14.0 release added the `routingTimeOverride` guard around `_DepartureTimeChip` in the mobile layout, but the wide-layout copy of the chip had no guard. Web users saw "Salir ahora" even when the override was set.

Mobile masked this because it never renders the wide layout.

Both copies now share the same guard.

### 2. `Provider<AppConfiguration?>` registration

Defensive cleanup: `MultiProvider` in `TrufiApp` registered the value as `Provider<AppConfiguration>` (non-nullable), but every consumer reads it with `context.watch<AppConfiguration?>()` (nullable). Provider package handles this gracefully via `Provider.of<T?>` falling back to null, so this isn't strictly needed — but registering the type explicitly nullable makes the contract obvious and matches the consumer side.

## Test plan

- [x] `melos run ci:analyze` clean
- [x] Local `flutter run -d chrome` — chip hidden, plan request goes out with `time=12:00`
- [x] Web build deployed to planner.trufi.app, verified in incognito — chip hidden, HH:mm labels suppressed in cards / list / detail

CHANGELOG: 5.14.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 5.14.1
+
+### Bug Fixes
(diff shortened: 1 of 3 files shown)


SamuelRioTz:
Closing — opening a clean PR with only the wide/web layout chip-guard fix (the Provider<AppConfiguration?> change in this branch turned out to be unnecessary).

PR #908, opened 2026-05-24
fix: Tu Ubicación button freeze and silent location fallback (#897)

### LINKED ISSUE #846: [Bug] [v5-Cochabamba] Error that leaves the screen blank when using "Current Location"
Bug description

The web application crashes and displays a completely white screen (White Screen of Death) immediately after interacting with the "Current Location" button. Interestingly, the feature worked correctly during the first interaction but failed consistently thereafter.

Medium: App freezes after clicking "Current Location"
Criticality: Major / High (Interrupts core navigation and map usage)

Test Environment
Device: Desktop PC / Laptop (Windows 10)
Browser: Chrome Version 145.0.7632.117 (Official Build) (64-bit)
Instance: v5-Cochabamba

Steps to Reproduce
Open the application in the browser: https://planner.trufi.app/
Click on "Select Origin" (Seleccionar origen).
Click on the "Your Location" (Tu Ubicación) button to retrieve the current position.
Observed Result: The screen immediately turns white, causing complete loss of UI and user confusion.

Current Result

The application fails to handle the state update after the first location request, leading to a "White Screen of Death" (WSOD). The entire user interface disappears, and the map fails to render, leaving the user with a blank browser window and no way to continue navigation.
 
Expected result

The application should successfully retrieve the user's coordinates and center the map on their current position without crashing the UI. If a location cannot be retrieved or a secondary click occurs, the app should either maintain the current view or show a non-intrusive error message (like a snackbar), but it should never result in a complete system freeze or white screen.
   
Attachments

Image

### LINKED ISSUE #897: [Bug] [v5-Cochabamba] UI freeze and unresponsiveness when clicking 'Tu Ubicación' button
**Bug description**

When clicking the 'Tu Ubicación' button, the application becomes unresponsive and the interface freezes. After several repeated attempts to press the button, the map finally displays the current location. This suggests a performance bottleneck or a race condition/error in the execution of the location event.

**Severity**
Criticality: High
Priority: High

**Environment**

- Device: Desktop PC / Laptop
- OS: Windows 10
- Browser: Chrome / Edge

**Steps to Reproduce**
1. Open the Trufi App.
2. Locate and press the 'Tu Ubicacion' button.
3. Attempt to interact with the map or other UI elements immediately after.
4. Observe the search results.

**Current Result**

The UI freezes completely upon pressing the button. The application remains stuck until, after multiple clicks, the location is finally displayed.

**Expected result**
The application should respond immediately to the click (e.g., showing a loading indicator) and smoothly move the map to the current location without blocking the main interface thread.
   
**Attachments**

https://github.com/user-attachments/assets/bbc08088-892f-4e25-b357-7a0c5b51a146

guss1215:
## Summary

Fixes the two distinct issues reported in #897:

- **Bug A — Web/Desktop freeze on the map's *Mi Ubicación* button.** `LocationService.startTracking()` could hang indefinitely waiting for a high-accuracy fix because its internal `getCurrentPosition()` had no `timeLimit`. Added a 10-second `timeLimit` and removed the redundant `getCurrentLocation()` call that `_onMyLocationPressed` made after `startTracking()` — the initial position is already captured in `LocationService.currentLocation`. Also added a `getLastKnownLocation()` fast-path so the camera repositions instantly while the accurate fix is acquired.
- **Bug B — Mobile silent fallback in the route search.** `onYourLocation` inside the search flow silently returned the city's `defaultCenter` for *every* failure case (permission denied forever, services disabled, timeout, null location), which appeared to users as a "random location" with no permission prompt. It now handles `deniedForever` / `serviceDisabled` with the same dialogs as the map button, and returns `null` (which keeps the user on the search screen) instead of teleporting them to `defaultCenter`. Also added a fast-path that reuses `currentLocation` when the map is already tracking.

### Other changes
- `LocationService` now accepts an optional `GeolocatorPlatform` via the constructor — production code path is unchanged (defaults to `GeolocatorPlatform.instance`), but the seam enables unit testing with `mocktail`.
- `LocationService.dispose()` is now idempotent.

### Tests
Added 25 unit tests in `packages/trufi_core_utils/test/location_service_test.dart` covering:
- Permission enum mapping (`always`, `whileInUse`, `denied`, `deniedForever`, `unableToDetermine`)
- `serviceDisabled` handling for both `checkPermission` and `requestPermission`
- `getCurrentLocation` / `getLastKnownLocation` (granted, denied, null, error paths)
- `startTracking` lifecycle, no-op-when-already-tracking, stream updates
- **Regression test pinning the 10-second `timeLimit` on the initial `getCurrentPosition` call (this is what fixes Bug A)**
- `stopTracking` / `dispose` lifecycle

Closes #897.

## Test plan

- [x] `flutter test packages/trufi_core_utils/test/location_service_test.dart` — 25 tests pass
- [x] `dart analyze` clean on all changed files
- [ ] Manual on a mobile device with the Cochabamba app:
  - [ ] Map *Mi Ubicación* button: camera jumps to last-known instantly, then refines. No multi-click "freeze".
  - [ ] Search field *Tu Ubicación* with permission never granted: system permission prompt appears.
  - [ ] Permission `deniedForever`: dialog with "Open Settings" (not a silent jump to defaultCenter).
  - [ ] Location services off: "Enable location" dialog.
  - [ ] Permission denied once, user cancels: stays on search screen (no random fallback).
  - [ ] Map already tracking → search *Tu Ubicación* reuses cached location without re-prompting.

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -594,15 +594,42 @@ class _HomeScreenState extends State<HomeScreen>
             chooseOnMapText: l10n.chooseOnMap,
           ),
           onYourLocation: () async {
-            // Try to get GPS location
+            // Fast path: reuse a fresh location when the map is already tracking.
+            if (_locationService.isTracking &&
+                _locationService.currentLocation != null) {
+              final loc = _locationService.currentLocation!;
+              return SearchLocation(
+                id: 'current_location',
+                displayName: l10n.yourLocation,
+                latitude: loc.latitude,
+                longitude: loc.longitude,
+              );
+            }
+
             var status = await _locationService.checkPermission();
 
             if (status == LocationPermissionStatus.denied) {
               status = await _locationService.requestPermission();
             }
 
-            if (status == LocationPermissionStatus.granted) {
-              final location = await _locationService.getCurrentLocation();
+            if (status == LocationPermissionStatus.deniedForever) {
+              if (mounted) _showPermissionDeniedDialog();
+              return null;
+            }
+
+            if (status == LocationPermissionStatus.serviceDisabled) {
+              if (mounted) _showLocationDisabledDialog();
+              return null;
+            }
+
+            if (status != LocationPermissionStatus.granted) {
+              return null;
+            }
+
+            try {
+              final location = await _locationService.getCurrentLocation(
+                timeout: const Duration(seconds: 10),
+              );
               if (location != null) {
                 return SearchLocation(
                   id: 'current_location',
@@ -611,15 +638,13 @@ class _HomeScreenState extends State<HomeScreen>
                   longitude: location.longitude,
                 );
               }
+            } catch (_) {
+              // Timeout or geolocator failure — fall through to null so the
+              // user stays on the search screen instead of being silently
+              // teleported to the city's default center.
             }
 
-            // Fallback to default center if GPS not available
-            return SearchLocation(
-              id: 'current_location',
-              displayName: l10n.yourLocation,
-              latitude: defaultCenter.latitude,
-              longitude: defaultCenter.longitude,
-            );
+            return null;
           },
           onChooseOnMap: () async {
             final result = await Navigator.push<MapLocationResult>(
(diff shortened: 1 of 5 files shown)


SamuelRioTz:
Thanks for the fix — the root-cause analysis on both bugs is solid and the regression test pinning the 10s `timeLimit` on the initial `getCurrentPosition` is exactly the kind of guard that will keep Bug A from coming back.

Before this can merge, could you complete the manual mobile checklist from the PR description and attach proof (a short screen recording is ideal, or screenshots per step)? The unit tests cover the service contract, but the user-visible flows still need device verification:

- [ ] Map *Mi Ubicación* button: camera jumps to last-known instantly, then refines. No multi-click freeze.
- [ ] Search field *Tu Ubicación* with permission never granted: the system permission prompt appears.
- [ ] Permission `deniedForever`: the "Open Settings" dialog shows (not a silent jump to `defaultCenter`).
- [ ] Location services off: the "Enable location" dialog shows.
- [ ] Permission denied once and the user cancels: stays on the search screen (no random fallback to `defaultCenter`).
- [ ] Map already tracking → search *Tu Ubicación* reuses the cached location without re-prompting.

Ideally the recording is taken on the Cochabamba build, since that is where #897 was reported. Thanks!

SamuelRioTz:
Thanks for the fix @guss1215 — the root cause analysis on both bugs is solid, and verifying it on a real Android device confirms the flow works as described: the system permission prompt now shows up correctly when tapping "Tu Ubicación" in the search, and denying or cancelling keeps the user on the search screen instead of teleporting them to `defaultCenter`.

I took the liberty of pushing one small follow-up commit (`00b739d7`) on top. Once the user grants permission, acquiring a fresh GPS fix can take several seconds, and during that window the "Tu Ubicación" tile stayed inert with no feedback — so I added a `CircularProgressIndicator` in place of the icon while the request is in flight, dimmed the label, and disabled the tap to prevent double-firing. The mounted checks around the async gap were also tightened to use `context.mounted` for the `BuildContext` use and `mounted` for the `setState`, which the analyzer was flagging.

Verified on a Pixel 9a (Android 16) over Tailscale ADB. Happy to revert the follow-up if you'd rather keep this PR scoped strictly to the bug fix.

SamuelRioTz (approved):

Issue #901, opened 2026-05-14
[Bug][v5-Cochabamba] Privacy permission revoke toggle is not updated instantly after interaction


jheimc:
### 📝 Description
When users revoke privacy permissions using the corresponding toggle/button, the UI does not update immediately after the interaction.  

From the user perspective, it appears as if nothing happened after pressing the option, since the permission state remains visually unchanged. The updated state is only reflected after navigating to another section of the app and returning to the privacy settings screen.

This behavior may confuse users and create uncertainty about whether the action was successfully completed.

* **Severity:** Low – Causes confusing and misleading UI feedback  
* **Priority:** Low – Permission changes should provide immediate visual confirmation  

### 📱 Test Environment
* **Device:** Mobile: Redmi Note 11 Pro+  
* **OS:** Android  
* **App Version:** Trufi v5 Cochabamba  

---

### 🛠 Steps to Reproduce
1. Open the **Trufi mobile app v5**  
2. Navigate in the sidebar to Settings > **Privacy** section  
3. In the share anonymous usage data Tap the option/button to revoke a privacy permission  
4. Observe that the toggle/state does not visually update immediately  
5. Navigate to another section of the app  
6. Return to the **Privacy Settings** screen  
7. Observe that the permission state is now updated  

### 📎 Evidence

https://github.com/user-attachments/assets/f7aadd67-af04-4dbf-86cd-6d0f30515c01

### ❌ Actual Behavior
The privacy permission toggle/button does not visually update immediately after the user revokes permissions, making it appear that the action was not applied.

### ✅ Expected Behavior
The privacy permission state should update instantly after the user interaction, providing immediate visual feedback that the permission was successfully revoked.

PR #910, opened 2026-05-26
fix: update privacy toggle to use ListenableBuilder for state management

jh0nz:
# Description
This PR fixes the UI bug #901 where the privacy permission revoke toggle did not visually update immediately after user interaction. 

## Root Cause
The `_PrivacySettingsCard` was a `StatelessWidget` that read `privacyConsentManager.isAccepted` during its initial build but lacked a mechanism to listen to reactive updates. Therefore, when `acceptConsent()` or `declineConsent()` triggered, the UI didn't re-render until the user forced a hard navigation back and forth.

## Solution
Wrapped the `_PrivacyToggleTile` inside a `ListenableBuilder` bound to `privacyConsentManager`. This ensures that whenever the consent manager dispatches a change notification, Flutter triggers a targeted re-render of the toggle widget instantly.

# Result
https://github.com/user-attachments/assets/aced732a-9624-46b4-827a-a2ee597236c3

### DIFF
--- a/apps/example/lib/main.dart
+++ b/apps/example/lib/main.dart
@@ -231,6 +231,20 @@ void main() {
           label: 'Instagram',
         ),
       ],
+      overlayManager: OverlayManager(
+        managers: [
+          OnboardingManager(
+            overlayBuilder: (onComplete) =>
+                OnboardingSheet(onComplete: onComplete),
+          ),
+          PrivacyConsentManager(
+            overlayBuilder: (onAccept, onDecline) => PrivacyConsentSheet(
+              onAccept: onAccept,
+              onDecline: onDecline,
+            ),
+          ),
+        ],
+      ),
       providers: [
         ChangeNotifierProvider(
           create: (_) => MapEngineManager(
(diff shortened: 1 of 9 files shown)


jh0nz:
@SamuelRioTz

SamuelRioTz:
Hey Jhon — thanks for the fix, the root-cause analysis was spot on.

Wanted to flag that the missing reactivity is actually surfacing a deeper bit of tech debt: child `AppOverlayManager`s were never registered as Providers, even though `AppOverlayManager`'s own docstring promises *"Automatically injected as providers by TrufiApp"* (lines 8-9 of `app_overlay_manager.dart`). That's why `settings_screen` has to reach into `OverlayManager` to look the consent manager up and layer `ListenableBuilder` on top — the canonical path (`context.watch<PrivacyConsentManager>()`) wasn't available.

I pushed a refactor on top of this branch that fulfills that promise:

- New `overlayManager` field on `AppConfiguration` (optional).
- `OverlayService` exposes a `providers` getter that returns itself + each child manager wrapped in a `ChangeNotifierProvider` of its concrete type, via a new `asProvider()` method on `AppOverlayManager`.
- `TrufiApp` spreads `...?config.overlayManager?.providers` into the root `MultiProvider`.
- Settings screen now reads the consent state in a single reactive line: `PrivacyConsentManager.maybeWatch(context)` — no `ListenableBuilder`, no manual lookup, no asymmetry with how `MapEngineManager` / `RoutingEngineManager` are already consumed elsewhere.

`flutter analyze` clean. Existing consumers that register `OverlayManager` inside the generic `providers` list need to migrate to the new dedicated field (already done in `apps/example`).

Happy to split this into a separate PR if it inflates the scope here too much — figured worth pairing because the toggle bug and the tech debt have the same root.

SamuelRioTz (approved):

Issue #904, opened 2026-05-15
[Bug] [v5-Cochabamba] Unintuitive POI interaction (long-press required) and missing location names


MelCampero11:
**Bug description**

Testing on the mobile APK revealed that interacting with Points of Interest (POI) icons is unintuitive as they do not respond to a single tap; users are forced to perform a long press to trigger the selection menu. Additionally, once a destination is selected, the specific name of the POI is replaced by raw geographic coordinates (e.g., -17.363...), causing a loss of context for the user compared to the Web version.

**Severity**
* Priority: High (Core navigation feature is difficult to discover)
* Criticality: High (Major) - Users cannot easily identify or confirm their destination by name, only by coordinates.

**Steps to Reproduce**
1. Open the Trufi Mobile App (APK).
2. Navigate the map and attempt to select a POI icon (e.g., Food, Shopping, or a specific landmark) with a single tap. Observe no response.
3. Perform a long press on the same POI icon to open the "Set as destination" (Establecer como destino) menu.
4. Select the point as a destination.
5. Observe the destination input field at the top and the route results panel.

**Actual Behavior**
- Interaction: Single taps on POI icons are ignored; only long presses trigger the selection menu.
- Metadata: The specific POI name (e.g., "Empanadas Pamelita") is replaced by generic coordinates (e.g., "Ubicación seleccionada, -17.36312...").
- Visuals: The route panel fails to display the descriptive name of the destination, showing only transit stops and coordinate strings.

**Expected Behavior**
- Single Tap Support: POIs should be selectable with a standard single tap/click.
- Name Persistence: The app should retrieve and display the descriptive POI name in the search bar and the route summary, maintaining parity with the Web version's behavior.
   
**Attachments**
* APK

<img width="576" height="1280" alt="Image" src="https://github.com/user-attachments/assets/f18b7e94-f40c-4c7f-bb85-2b15f4739aa3" />

<img width="576" height="1280" alt="Image" src="https://github.com/user-attachments/assets/29e6a5d1-1f3b-404e-b3ed-8a33865f1849" />

**Environment**
 - Device: Mobile (Xiaomi Redmi Note 9 Pro)
 - OS: Android 11
 - App Version: Trufi v5 Cochabamba (APK)

MelCampero11:
**Update: Cross-platform Interaction Issue (Web & Mobile) and Metadata Loss**

[...]

PR #916, opened 2026-06-17
fix: reverse-geocode map-picked locations to show place names (#904)

guss1215:
## What & why

When a user picks an arbitrary point on the map — via **long-press** or **"Choose on map"** — the resulting origin/destination was built with a generic label (`l10n.selectedLocation`) and the **raw coordinates** as the address. The search bar and route summary therefore showed e.g. `-17.36369, -66.14936` instead of a real place name.

`SearchLocationsCubit.reverseGeocode()` (backed by `PhotonSearchService`/`NominatimSearchService`) already existed end-to-end but was **never called** from the map interaction handlers. POI taps already used `poi.displayName`; only arbitrary map-picked points fell back to coordinates.

This addresses the **"missing location names"** part of #904.

## Changes

- Add a shared `_resolvePickedLocation(lat, lng)` helper in `home_screen.dart` that reverse-geocodes the picked point through `SearchLocationsCubit`:
  - 5s timeout so route planning is never blocked by a slow geocoder.
  - Graceful fallback to the previous generic label (coordinates kept only in `address`) when the cubit is unavailable, the lookup times out, or it returns nothing.
- Use the helper in both map-pick flows: `_onMapLongPress` (set origin/destination) and the "Choose on map" result builder.

Single file changed: `packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart`.

## Verification

- `flutter analyze` → no issues.
- Ran `apps/example` and picked a point (`-17.39884, -66.16269`) via "Choose on map": the destination field now shows **"Pasaje A, Cochabamba"** (reverse-geocoded) instead of the raw coordinates.

## Out of scope / notes

- The **"single tap on POI is ignored"** part of #904 is a separate, deployment-level matter (POI categories are disabled by default in the app's `metadata.json`, so there are no interactive markers to tap) and is **not** included here.
- Optional follow-up: pass `language: 'es'` to `PhotonSearchService` in the app for localized reverse-geocode names.


SamuelRioTz (approved):
**Verdict: ✅ Approved** — re-reviewed after all in-review changes (`584495bf` main merge, `321e0a1e` guard+feedback, `9f488040` per-field guard fix)

## Code review

- Solid core: the reverse-geocode goes through `SearchLocationsCubit` with a 5 s timeout and full fallback chain (timeout, failure, empty result, missing cubit → generic label with raw coordinates), so route planning is never blocked by the lookup.
- The June patch aged well: it merged onto today's `main` (12 PRs later) with zero conflicts.
- Review asks fixed in-review (`321e0a1e`), then audited by a second re-analysis pass which caught a bug in the first guard (`9f488040`):
  1. **Race guard, per field**: origin and destination advance independent tokens — two quick long-presses on the same field no longer race, while an origin pick followed by a quick destination pick keeps **both** (a single global token silently dropped the origin; verified live below).
  2. **Progress feedback**: a transient snackbar with a spinner covers the up-to-5 s lookup (new `resolvingLocation` l10n key, en/es/de) instead of the UI appearing frozen.
  3. `address` falls back to the raw coordinates when the geocoder returns none.
- Tests: home_screen suite 7/7 green locally (CI does not run tests); analyze clean (the one `info` is pre-existing).

| Long-press sheet | Lookup in progress | Resolved to a real place name |
|---|---|---|
|  |  |  |

Long-pressed an arbitrary map point → "Obteniendo el nombre del lugar…" → destination shows **"Calle Ladislao Cabrera, Cochabamba…"** instead of `-17.39872, -66.16279`.

**Crossed-picks regression test** (the case the per-field fix protects): origin picked, destination picked within the resolve window — both fields kept their reverse-geocoded names:



## Follow-ups (non-blocking)

- Choose-on-map flow: silent up-to-5 s wait on the search screen (modal flow, no shared state — correctly tokenless) and a `context` use without `mounted` guard; same pattern predates this PR.

Issue #929, opened 2026-07-16
[Bug][v5-Cochabamba] Route planner returns significantly fewer route alternatives


jheimc:
### 📝 Description
Compared to previous versions, the route planner now returns considerably fewer route alternatives for the same origin and destination. This limits users' ability to choose the most convenient journey and may hide valid transport options.

* **Severity:** Medium – Reduces routing flexibility.
* **Priority:** Medium – Impacts route selection but does not prevent route planning.

### 📱 Test Environment
* **Device:** Mobile: Redmi Note 11 Pro+
* **Device:** PC / Laptop
* **OS:** Windows 11
* **Browser:** Chrome Version 145.0.7632.109 (Official Build) (64-bit)

---

### 🛠 Steps to Reproduce
1. Open **Trufi App v5**.
2. Enter a valid origin and destination.
3. Search for available routes.
4. Compare the number of suggested routes with previous application versions.

### 📎 Evidence
(User reports indicating fewer route alternatives are returned after the latest update.)

<img width="1510" height="305" alt="Image" src="https://github.com/user-attachments/assets/b1d11cb0-4195-4545-a93f-4b08a569cad2" />

### ❌ Actual Behavior
The application returns significantly fewer route alternatives than expected.

### ✅ Expected Behavior
The route planner should return all relevant route alternatives available for the selected trip.

MaiteAraucoPorrez:
**Case 1: Route planner ignores existing direct trufi lines and forces unnecessary transfers**

* Current result: For origin-destination pairs where multiple trufi lines already pass through the same street/point and reach the destination directly, the app forces the user through 2-3 transfer trufis instead of offering the direct line(s) that exist on that same route.

* Expected result: When multiple trufi lines cover the same corridor as the requested destination, the planner should surface those direct options first, before suggesting multi-leg alternatives.

* Evidence:
User review report (July 2026), "Ismael Aleluya Misericordia" (9 jul): reports being routed through 2-3 trufis "cuando hay trufis que pasan por la misma ruta y te llevan al mismo punto."

<img width="852" height="115" alt="Image" src="https://github.com/user-attachments/assets/2df7d8d4-3caf-451a-bb44-a7911395e6f5" />

<img width="612" height="1299" alt="Image" src="https://github.com/user-attachments/assets/7c611eb2-6e85-4915-8a9e-2f1dd89972a3" />

[...]

PR #932, opened 2026-07-20
feat(routing): make the itinerary count app-configurable

### LINKED ISSUE #923: [Feature][v5-Cochabamba] Increase maximum routes displayed from 5 to 10
**Actual result**

Users are requesting more route options in the search results. Currently, the application displays a maximum of 5 routes in the search results, but we are receiving multiple complaints from users indicating that this limit is too restrictive. Users would benefit from having access to more alternative routes to choose from, which would improve their trip planning experience and satisfaction.

**Expected**

Increase the maximum number of routes displayed in the search results from 5 to 10. This would provide users with more route alternatives without overwhelming the UI, allowing them to make more informed decisions when planning their trips.

**Additional context**

- Affects both web and mobile (v5) versions of the application
- Users have expressed dissatisfaction with the limited route options
- This is a quick win for improving user experience and addressing user feedback

guss1215:
Reworked after review: **trufi-core defaults stay untouched** — this PR only adds the override point plus the plumbing to honor it end-to-end. City apps opt in (Cochabamba will set `maxItineraries: 20` in trufi-app); every other Trufi keeps getting 5.

## Changes

- `RoutingEngineManager` gains **`maxItineraries`** (constructor param, default **5** = historical behavior). `fetchPlan` uses it when the caller doesn't pass an explicit `numItineraries`.
- `RemotePlannerClient` now sends `maxResults` in the `/plan` body so the override reaches trufi-server-planner (which clamps it server-side; older deployments ignore the extra field).
- `LocalPlannerClient` scales the GTFS service per-bucket caps (`maxDirects`/`maxTransferPaths`) with `maxResults` — without this, an app-level override gets silently clipped to 5 directs + 5 transfers. At the default of 5 the behavior is identical to before.
- No defaults changed in providers or `GtfsRoutingService`.

## Verified in trufi-app (Cochabamba, offline GTFS, Android emulator)

Same three dense pairs, only difference is the app passing `maxItineraries: 20`:

| Pair | stock app (default 5) | with app override 20 |
|---|---|---|
| Centro → Av. América | 5 (capped) | **14** |
| Centro → Quillacollo | 5 (capped) | **20** |
| Cala Cala → La Cancha | 5 (capped) | **18** |

(`TrufiPlannerProvider: Found N paths` in logcat; the control run proves the default path is untouched.)

Unit tests: default stays 5, app override honored, explicit `numItineraries` wins (4/4).

Closes #923. Related: #929, trufi-server-planner#4.

### DIFF
--- a/packages/trufi_core_planner/lib/src/client/local_planner_client.dart
+++ b/packages/trufi_core_planner/lib/src/client/local_planner_client.dart
@@ -79,6 +79,12 @@ class LocalPlannerClient implements PlannerRoutingClient {
       destination: destination,
       maxWalkDistance: maxWalkDistance,
       maxResults: maxResults,
+      // Scale the per-bucket caps with the requested result count so an
+      // app-level maxItineraries override isn't silently clipped to the
+      // service defaults (5 directs + 5 transfers). At the default
+      // maxResults = 5 this is identical to the previous behavior.
+      maxDirects: maxResults,
+      maxTransferPaths: maxResults,
     );
   }
 
--- a/packages/trufi_core_planner/lib/src/client/remote_planner_client.dart
+++ b/packages/trufi_core_planner/lib/src/client/remote_planner_client.dart
@@ -69,6 +69,10 @@ class RemotePlannerClient implements PlannerRoutingClient {
       body: jsonEncode({
         'from': {'lat': origin.latitude, 'lon': origin.longitude},
         'to': {'lat': destination.latitude, 'lon': destination.longitude},
+        // Forward the requested result count so an app-level override
+        // reaches the server (older servers ignore unknown fields; the
+        // planner server clamps it to its own limits).
+        'maxResults': maxResults,
       }),
     );
 
--- a/packages/trufi_core_routing/lib/src/routing_engine_manager.dart
+++ b/packages/trufi_core_routing/lib/src/routing_engine_manager.dart
@@ -43,14 +43,27 @@ class RoutingEngineManager extends ChangeNotifier {
   int _currentIndex;
   bool _isPreloading = false;
 
+  /// Maximum number of itineraries requested per plan when [fetchPlan] is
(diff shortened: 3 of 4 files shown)


SamuelRioTz (approved):
**Verdict: ✅ Approved**

## Code review

- Config injected properly through the constructor chain (manager → provider → local/remote client) — no mutable global state; the screens layer calls `fetchPlan` without `numItineraries`, so the app-level override genuinely governs.
- Historical default of 5 preserved and pinned by a dedicated test; explicit `numItineraries` wins over the app override. Not a breaking change for city apps.
- Trivial textual overlap with #934 in `remote_planner_client.dart` (both insert one line in the `/plan` body) — 1-line rebase for whichever lands second.
- Minor notes: the `assert` guard doesn't protect release builds; the remote override depends on trufi-server-planner#4.

## Verification (this branch only)

Unit tests run locally (CI does not run tests) — all four pass:

```
RoutingEngineManager fetchPlan keeps the historical default of 5 itineraries
RoutingEngineManager fetchPlan forwards an explicit numItineraries unchanged
RoutingEngineManager fetchPlan uses the app-level maxItineraries override as the default (issue #923: Cochabamba requests 20)
RoutingEngineManager fetchPlan an explicit numItineraries wins over maxItineraries
00:00 +4: All tests passed!
```

Android emulator, Trufi Planner, end-to-end on both sides of the feature:

1. **Default (no override configured):** cap of 5 respected — 4 viable options for a 1.1 km trip (≤ 5); longer routes return exactly 5.
2. **Override exercised for real:** temporarily wired `RoutingEngineManager(engines: ..., maxItineraries: 10)` in the example app and replanned the same Plaza Libertad → Terminal de Buses trip that previously capped at 5 — **"10 rutas encontradas"**.

| Default: ≤5 results | Override 10: "10 rutas encontradas" |
|---|---|
|  |  |

PR #933, opened 2026-07-20
feat: open the routes list filtered by operator via deep link

### LINKED ISSUE #924: [Feature] [v5-Cochabamba] Generate QR Code to Open Routes Filtered by Operator Description
As part of collaboration with public transport operators, we would like to have the ability to generate QR codes that open the Trufi App directly to the Routes section, displaying only the routes belonging to a specific transport operator.

This would allow passengers to scan a QR code placed inside vehicles or at bus stops and immediately access the relevant routes without manually searching through all available lines.

**Proposed Solution**

Implement a deep link (or similar mechanism) that supports the operator field as a filter.

Example flow:

1. User scans the QR code.
2. The Trufi App opens automatically.
3. The app navigates to the Routes section.
4. Only the routes associated with the specified operator are displayed.

Example (concept only):

trufi://routes?operator=operator_name

or

https://trufi.app/routes?operator=operator_name

The exact implementation can follow the application's existing navigation architecture.

**Expected Result**

After scanning the QR code:

- The Trufi App opens automatically.
- The Routes screen is displayed.
- Only the routes belonging to the specified operator are shown.
- Users can select one of those routes without browsing the complete route list.

guss1215:
Fixes #924

## What

Adds end-to-end support for QR codes that open the app directly on the Routes screen showing only one operator's lines — **including generating the QR from inside the app**.

### Deep link + filter

```
trufiapp://routes?operator=Sindicato%20Pucara%20Grande
https://<app-host>/routes?operator=Sindicato%20Pucara%20Grande
```

- **transport_list** — first-class agency filter on `TransportListDataProvider` (`setAgencyFilter`, case-insensitive exact match on `agencyName`, intersects with the text search; `reapplyFilters` restores scoping after `load`/`refresh` reset the list). The screen seeds it from the `operator` query parameter and shows a dismissible chip; dismissing it also drops the parameter from the URL.
- **trufi_core_ui** — `DeepLinkService` now forwards links that are not shared-route links as in-app locations (`resolveInAppLocation` maps custom-scheme hosts to the first path segment), and `TrufiApp` navigates to them when they target a registered screen; unknown links keep being ignored.

### In-app QR generation (second commit)

A QR button on **every agency section header** and **next to the active operator chip** opens a dialog with the scannable code (white backdrop + quiet zone so it survives printing and dark mode), the encoded link, and Copy/Share actions (labels from `MaterialLocalizations` — no new ARB strings). Operators can share/print it straight from the app.

Links prefer the new `TransportListTrufiScreen.shareBaseUrl` (mirrors `HomeScreenConfig.shareBaseUrl`) and fall back to the app's `deepLinkScheme`; with neither configured the QR affordances are hidden. New deps: `qr_flutter`, `share_plus` (already used by home_screen).

Note: `QrImageView` must sit in a fixed-size `SizedBox` — it uses a `LayoutBuilder`, which cannot answer `AlertDialog`'s intrinsic-dimension queries; without it the dialog silently renders as scrim-only.

## Heads-up for app manifests

Flutter's built-in deep linking (default-on since 3.27) must be disabled when using this: the engine otherwise also pushes the raw URI into go_router, which clobbers the in-app navigation with a fallback to home. The example app's manifest shows the required setup (intent-filter host + meta-data); Cochabamba's app needs the same two lines to consume this feature.

## Verified on device (Cochabamba OTP data)

- Cold start via link: opens the filtered list — 21 routes for *Sindicato de taxis y trufis 14 de septiembre*, correctly matching both capitalization variants present in the GTFS thanks to the case-insensitive match.
- Warm link while the app runs: filter switches live to the new operator.
- Dismissing the chip: full list restored (279 routes), `operator` dropped from the URL.
- **Full scan round trip**: the in-app QR for *Sindicato Pucara Grande* decodes (OpenCV) to `trufiapp://routes?operator=Sindicato+Pucara+Grande`, and firing that decoded URL reopens the filtered list.
- Web already works via the router's `initialLocation`.

## Tests

- `agency_filter_test.dart`: exact/case-insensitive matching, unknown operator → empty, clear semantics, intersection with text query, re-scoping after refresh.
- `deep_link_location_test.dart`: URI → in-app location mapping for custom-scheme and https links.
- `operator_qr_dialog_test.dart`: dialog renders the QR, title, and link (guards the AlertDialog/LayoutBuilder regression).


SamuelRioTz (approved):
**Verdict: ✅ Approved** — re-reviewed end to end after all in-review changes

## Scope of the final state

Original feature (operator deep link + filter + QR sharing) plus six in-review commits: review fixes (`1f4e319b`), conflict resolution with the merged l10n series (`6d6ff8e1`), the operator QR reworked as a Trufi bottom sheet with inline link copy and a copy-QR-image action (`2fdf7403`, `6a1608b5`, `01abdd6e`), and lockfile sync (`fa2d30f3`).

## Code review (full re-analysis of the final diff)

- Deep-link plumbing: conservative parsing, navigation restricted to registered screens, dot-segment invariant pinned by test; unknown hosts rejected at the Android manifest level.
- Operator filter: case-insensitive exact match (groups GTFS spelling variants), intersects with search, survives refresh; public API fully additive.
- Conflict resolution audited: 29 l10n keys × 3 locales, no duplicates, generated files in sync; localized "Others" label merged with agency-aware headers.
- QR sheet audited: no leaks or async gaps (`mounted` guarded), print-quality PNG (white background, ~4-module quiet zone), copy failures surface a localized snackbar instead of an unhandled `PlatformException`, rasterized image disposed.
- New dependency `pasteboard` scoped to the screen package; host apps need a FileProvider (**cache-path**, not the external-path the package README suggests) — wired in the example as the reference.
- Tests: 18 green locally (CI does not run tests); analyze clean.

## Runtime verification (fresh, final build)

| Android: deep link → filtered list | QR sheet (final design) | Copy QR → system clipboard overlay | iOS: same deep link |
|---|---|---|---|
|  |  |  |  |

Zero exceptions in logcat across the flows. Side effect of the main merge: the init splash now localizes on this branch too.

## Follow-ups (non-blocking)

- `pasteboard` 0.5.0 does not await the browser clipboard promise on web — the success check can lie on Safari; upstream issue.
- `shareBaseUrl` drops subpaths (same pre-existing debt as home_screen); iPad share sheet lacks `sharePositionOrigin` (repo-wide debt).
- Adopting city apps need: the two AndroidManifest lines, the Info.plist key, and the FileProvider entry — all shown in the example app.

PR #934, opened 2026-07-21
feat(planner): walk a bit farther to reach direct lines

### LINKED ISSUE #926: [Bug][v5-Cochabamba] Route planner generates inefficient routes by prioritizing unnecessary transfers over short walking segments
### 📝 Description
After the latest update, the route planning algorithm generates inefficient itineraries by prioritizing multiple public transport transfers instead of considering short walking distances to reach a direct line.

For trips where users could walk a reasonable distance (e.g., 1–10 blocks) and board a single trufi, the application instead recommends routes requiring two or more transfers. This results in longer, more complex journeys and significantly reduces the quality of the generated routes compared to previous versions.

The routing algorithm should balance walking distance and transfers to provide the most efficient journey, minimizing unnecessary transfers whenever a direct route can be reached with a reasonable walk.

* **Severity:** High – Users receive inefficient routes that increase travel time and complexity.
* **Priority:** High – This directly impacts the application's primary functionality.

### 📱 Test Environment

---

### 🛠 Steps to Reproduce
1. Open **Trufi App v5**.
2. Select an origin located near multiple public transport lines.
3. Select a destination that can be reached by walking a short distance to a direct trufi line.
4. Search for available routes.
5. Review the suggested itineraries.

### 📎 Evidence
(User reports indicating the planner recommends multiple transfers instead of allowing a short walk to access a direct route.)

Image

Image

Image

Image

### ❌ Actual Behavior
The planner prioritizes routes with multiple transfers and rarely considers reasonable walking distances to access more efficient direct routes.

### ✅ Expected Behavior
The routing algorithm should evaluate reasonable walking distances together with available public transport options and prioritize the most efficient journey by:
- Minimizing unnecessary transfers.
- Suggesting short walking segments when they lead to a simpler direct route.
- Returning routes that optimize the overall travel experience rather than simply minimizing walking distance.

guss1215:
Fixes #926

## The problem

Users get transfer-heavy itineraries even when a direct line is boardable a few blocks away. Two planner limits made that line invisible:

1. **Only the 10 nearest stops** were considered as boarding/alighting candidates. Cochabamba's network is dense — the direct line's stop is often just slightly farther than the closest cluster, so it never entered the search and only transfer combinations survived.
2. **The walk radius was hardcoded to 500 m** at the provider call site. `TrufiPlannerConfig.maxWalkingDistance` already existed as a config knob, but it was never wired — the value was silently ignored.

## The fix

- `GtfsRoutingService.findRoutes` takes `maxStopCandidates` (default 60) and considers every stop within the walk radius, not just the nearest 10.
- `TrufiPlannerProvider` now honors `config.maxWalkingDistance`, and the default rises to **800 m (~8 blocks)** — walking a few extra blocks to a direct ride beats a transfer, which is exactly what the reports ask for. Apps can tune it per deployment via `TrufiPlannerConfig`.
- `RemotePlannerClient` sends `maxWalkDistance` in the request body so the hosted planner applies the same radius (companion server change in trufi-association/trufi-server-planner#4).

## Measured on the real Cochabamba GTFS

300 random origin/destination pairs (≥ 1.5 km), same seed:

| configuration | pairs with results | direct line as top result |
|---|---|---|
| old (cap 10, 500 m) | 94 | 7 |
| cap 60, 500 m | 99 | 19 |
| **cap 60, 800 m (new defaults)** | **147 (+56%)** | **33 (4.7×)** |

The stop-candidate fix alone rescues 12 pairs from transfer-only/none to direct; the wider radius rescues 14 more. Search time stays ~0.8 ms per query.

## Tests

`walk_to_direct_routes_test.dart` builds a synthetic network with a direct line 600 m away behind a ring of 12 closer transfer-only stops, and pins all three behaviors: new defaults surface the direct line; the old stop cap alone hides it; the old 500 m radius alone hides it.

Note: `gtfs_routing_service_test.dart` (#859 scoring) fails identically on `main` — pre-existing, unrelated.

## Relationship to other PRs

Independent of #932 (itinerary count) — both touch `findRoutes` defaults, so whichever lands second needs a trivial rebase.


SamuelRioTz (approved):
**Verdict: ✅ Approved** (default change to 800 m confirmed by maintainer; review nit fixed in `5a60be60`)

## Code review

- The diagnosis is real (verified on `main`): `maxWalkDistance: 500.0` was hardcoded, silently ignoring `config.maxWalkingDistance`, and the candidate search capped at a handful of nearest stops — both together hid direct lines a short walk away (issue #926).
- The fix wires `config.maxWalkingDistance` through (default now 800 m) and widens the candidate pool (`maxStopCandidates = 60`). Only affects the TrufiPlanner (GTFS local/remote) — none of the three OTP backends.
- Includes 3 regression tests that isolate each old failure mode; all pass.
- Note: the pre-existing red test on `main` (`gtfs_routing_service_test.dart`, issue #859 scoring) is a *different* bug and fails identically with and without this PR — not a regression, tracked separately.
- Applies cleanly on current `main` (0 conflicts; the expected 1-line overlap with #932 resolved on its own).
- Review nit fixed in-review (`5a60be60`): `maxStopCandidates` is now exposed in `TrufiPlannerConfig` and threaded through provider → client → service (the stale 500 m defaults in client signatures were aligned to 800 m as well — call sites always pass explicit values, no behavior change). Verified end-to-end: default 60 → 6 routes; `maxStopCandidates: 1` wired in the app → transit collapses to the walk fallback.
- Remaining follow-ups (pre-existing, out of this PR's scope): `walkSpeed`/`maxTransfers` config fields are still unused; some `dart format` noise in the diff.

## Runtime verification (Android emulator, this branch only — both sides of the change)

| Default 800 m: 6 routes | `maxWalkingDistance: 60`: transit gone | `maxStopCandidates: 1` (new knob): walk fallback only |
|---|---|---|
|  |  |  |

Same trip (Tu ubicación → Terminal de Buses, Trufi Planner): the config value now genuinely governs stop reachability — with an absurd 60 m limit every transit option disappears and only the direct walk remains. Before this PR the config was dead and this experiment would have changed nothing.

## Product note

Default walking radius changes 500 m → 800 m for **all** apps using TrufiPlanner: more direct-line results and fewer "no routes" dead ends, at the cost of offering longer walks. Apps can pin the old behavior with `TrufiPlannerConfig(maxWalkingDistance: 500)`.

PR #935, opened 2026-07-21
fix(otp28): keep long walking routes when walk-only mode is selected

### LINKED ISSUE #900: [Bug][v5-Cochabamba] OTP 2.8.1 does not return walking routes for long-distance trips
### 📝 Description
When selecting just the **“Walk”** transportation modes for relatively long-distance routes, the application using **OTP version 2.8.1** does not return any route results.  

However, the same origin and destination combination successfully returns walking routes when using **OTP version 1.5.0**.

This creates inconsistent routing behavior between OTP versions and prevents users from obtaining valid walking directions for longer trips.

* **Severity:** Medium – Users are unable to retrieve valid walking routes in OTP 2.8.1  
* **Priority:** Medium – Routing consistency between OTP versions should be maintained  

### 📱 Test Environment
* **OS:** Android  
* **App Version:** Trufi v5 Cochabamba  
* **OTP Version Affected:** 2.8.1  
* **OTP Version Working:** 1.5.0  

---

### 🛠 Steps to Reproduce
1. Open the **Trufi mobile app v5**  
2. Select an origin and destination with a relatively long walking distance or open the following route: https://planner.trufi.app/?from_lat=-17.371319&from_lng=-66.175818&from_name=Selected+Location&to_lat=-17.394924&to_lng=-66.158313&to_name=Selected+Location&time_mode=leaveNow&route=0 
3. Wait until the app loads the routes
4. Go to route Settings button and select only the "Walk" option in Transport mode (uncheck any other option)
5.  Execute the route search using **OTP 2.8.1**  
6. Observe that no routes are returned  
7. Repeat the same search using **OTP 1.5.0**  
8. Observe that walking routes are successfully generated  
 
### 📎 Evidence
OTP 2.8.1 results: 

Image

OTP 1.5.0 results:

Image

### ❌ Actual Behavior
OTP 2.8.1 does not return walking routes for long-distance trips. 

### ✅ Expected Behavior
OTP 2.8.1 should return valid walking routes for long-distance trips, we should have the single route suggested no matter if it's a long or short route and be consistent with the behavior observed in OTP 1.5.0.

guss1215:
Fixes #900

## Root cause — not an OTP problem

The report blamed OTP 2.8.1, but the server answers correctly: the exact origin/destination from the issue returns a 3.9 km walking itinerary when queried directly (verified against `otp281.trufi.app` with the app's own GraphQL query and variables, byte-for-byte).

The bug is client-side: the OTP 2.8 provider **filters out walk-only itineraries longer than 1 km** after parsing. That's sensible noise reduction for *transit* searches — OTP returns a giant walk itinerary when no service is available, which only adds noise. But the filter also ran when the rider explicitly restricted transport modes to **Walk only**, where every result is by definition a walk — so any walk-only search beyond 1 km showed "no routes" while the server had answered fine.

Instrumented on-device confirmation: `raw itineraries: 1` from the server → `Itineraries found: 0` after the filter.

OTP 1.5 has no such filter, which is why it *appeared* to work there — the inconsistency between versions was entirely this one line.

## Fix

Skip the filter when the requested modes are walk-only. The transit-search behavior is unchanged (long walk noise still dropped). Extracted the filter into a static `filterLongWalks` for testability.

## Verified on device

With the report's coordinates, OTP 2.8.1 engine + Walk only: the app now shows the **52-minute, 4.0 km walking route** (before: no results).

## Tests

`otp_28_walk_filter_test.dart`: long walks kept in walk-only mode, still dropped from transit searches, short walks always survive, null passthrough.

### DIFF
--- a/packages/trufi_core_routing/lib/src/providers/otp_2_8/otp_28_routing_provider.dart
+++ b/packages/trufi_core_routing/lib/src/providers/otp_2_8/otp_28_routing_provider.dart
@@ -5,6 +5,7 @@ import 'package:graphql/client.dart';
 import 'package:trufi_core_interfaces/trufi_core_interfaces.dart';
 
 import '../otp_2_4/graphql_client_factory.dart';
+import '../../models/itinerary.dart';
 import '../../models/plan.dart';
 import '../../models/routing_location.dart';
 import '../../models/stop.dart';
@@ -110,8 +111,11 @@ class Otp28RoutingProvider extends IRoutingProvider {
   bool get requiresInternet => true;
 
   @override
-  Widget? buildPreferencesUI(BuildContext context) =>
-      Otp28Preferences(state: _prefs, showWheelchair: showWheelchairOption, showBicycle: showBicycleOption);
+  Widget? buildPreferencesUI(BuildContext context) => Otp28Preferences(
+    state: _prefs,
+    showWheelchair: showWheelchairOption,
+    showBicycle: showBicycleOption,
+  );
 
   @override
   void resetPreferences() => _prefs.reset();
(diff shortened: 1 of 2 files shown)


SamuelRioTz (approved):
**Verdict: ✅ Approved**

## Code review

- Surgical fix: the 1 km itinerary filter is skipped only when `transportModes == {walk}` on a non-simple query — verified case by case that transit, simple-query and mixed-mode requests produce byte-identical results to `main`.
- The null passthrough is equivalent to the previous behavior; only OTP 2.8 had this filter (1.5/2.4 don't need the fix).
- Includes regression tests; full package suite run locally (CI does not run tests) — no regressions vs `main`, new tests pass.
- No file overlaps with other open PRs (zero interaction with #934, which targets the offline TrufiPlanner, not OTP 2.8).

## Runtime verification (Android emulator, this branch only, live OTP 2.8 backend)

Planned "Tu ubicación → Terminal de Buses" (~1.1 km) on OTP 2.8.1, then switched transport modes to walk-only:

| Transit itineraries (baseline behavior) | Walk-only prefs | Walk-only result — the fix |
|---|---|---|
|  |  |  |

Walk-only over 1 km now returns "1 ruta encontrada" (1.1 km, 15 min walk) — before this fix the >1 km filter silently dropped it and the user saw no routes.

Issue #871, opened 2026-04-03
[Bug] [v5-Cochabamba] Inconsistent language translation when selecting German language


OriMads:
**Bug description**

When selecting German as the application language in the settings, the interface shows inconsistent translations.

Some elements are displayed in German, while others remain in English or Spanish. This results in a mixed-language interface, which affects usability and user experience.

**Severity**

:white_large_square: Low
:white_check_mark: Medium
:white_large_square: High

**Steps to Reproduce**

Open the application.
Go to the Settings section.
Change the language to German.
Navigate through different sections of the app.
Observe the language used in the interface.

**Current Result**

The interface displays mixed languages. Some fields are translated into German, while others remain in English or Spanish.

**Expected result**

All interface elements should be consistently translated into German when that language is selected.

**Attachments**

<img width="523" height="891" alt="Image" src="https://github.com/user-attachments/assets/2a7d238c-604f-46e5-84df-3af1dffc2555" />

<img width="786" height="568" alt="Image" src="https://github.com/user-attachments/assets/38d49915-a54a-44e6-8399-7bc69929c8cd" />

<img width="1845" height="897" alt="Image" src="https://github.com/user-attachments/assets/1b3e9eab-17c8-431e-a5ed-b0c44ef80342" />

<img width="1845" height="897" alt="Image" src="https://github.com/user-attachments/assets/fe0c24b8-04ba-4922-ac2a-0bbb002b9985" />

ux-ventura:
Esto es realmente urgente ya que no solo es con el idioma aleman sino tambien con los demas ya que existe una mezcla de spanglish

IsmaDavidAA:
<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/de22a584-6b3d-48cd-b6e8-4c94de0e849f" />

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/7da61148-c98f-4eac-afdd-1e2ee82612b3" />

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/8f9c658e-bb80-4c7f-ba7b-443386978961" />

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/b28cfe93-31c9-4eea-ada7-28f12774f850" />

guss1215:
[...]

PR #938, opened 2026-07-22
fix(saved_places): localize hardcoded strings (#871)

guss1215:
Part of #871 — saved_places slice.

Fixes the 6 hardcoded strings in this package. Interesting find: the four tile literals (`'Home'`, `'Work'`, `'Edit'`, `'Remove'`) were dead English fallbacks — `SavedPlacesLocalizations.of(context)` is non-nullable (`nullable-getter: false`) but two private members were typed nullable, keeping `?? 'Home'`-style fallbacks alive. Made those private members non-nullable and reused the existing `home`/`work`/`editPlace`/`removePlace` keys, so rendered text is unchanged.

- New keys `errorLoadingPlaces` and `retry` in en/es/de replace the `languageCode == 'es'` ternaries in the list error state.
- No public API changes (all edited members are library-private).

Known follow-up (pre-existing, outside this slice): `SavedPlacesCubit` stores raw `e.toString()` in `state.errorMessage` and the list shows it verbatim on real load failures — that path still surfaces an unlocalized exception string.

Verification: Translation Check replicated locally (clean tree after `melos run l10n --no-select`), `flutter analyze` clean, 10/10 package tests pass.

### DIFF
--- a/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
+++ b/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
@@ -31,5 +31,7 @@
   "locationRequired": "Standort ist erforderlich",
   "locationSelected": "Standort ausgewählt",
   "tapToSelectLocation": "Tippen, um auf der Karte auszuwählen",
-  "change": "Ändern"
+  "change": "Ändern",
+  "errorLoadingPlaces": "Fehler beim Laden der Orte",
+  "retry": "Erneut versuchen"
 }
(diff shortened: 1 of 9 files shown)


SamuelRioTz (approved):

SamuelRioTz:
**Verdict: ✅ Approved**

## Code review

- All 10 hardcoded strings correctly migrated to the package's existing l10n infrastructure (`saved_places_{en,es,de}.arb`); generated files in sync.
- Correct diagnosis of dead fallback code: with `nullable-getter: false` the null-checks were unreachable (verified against `l10n.yaml`).
- No public API changes, no file overlaps with any other open PR, no behavior changes in es/en (visual no-op there; fixes the German leak).

Built `apps/example` from this branch, verified Saved Places screen in all 3 languages:

| Cold start + planning OK | Spanish | German — no leaks |
|---|---|---|
|  |  |  |

## Non-blocking follow-up (already declared in PR description)

- Cubit still surfaces raw `e.toString()` in one error path — fine to address separately.

PR #936, opened 2026-07-21
fix(home): localize hardcoded strings that leaked into the German UI

guss1215:
First slice of #871.

## Root cause of the mixed-language UI

The ARB files are healthy — an audit found **zero missing keys** in `de`/`es` across all 11 localized packages (the handful of identical values are legitimate German: *Online, Details, System, Feedback…*). The leaks come from **~106 hardcoded string literals in widget code**, in three flavors:

1. `languageCode == 'es' ? 'Spanish' : 'English'` conditionals — German falls through to English.
2. Pure Spanish or English literals in `Text(...)` — shown verbatim in every language.
3. Whole screens with no l10n at all (OTP 1.5/2.4 preference sheets, search screen, map settings).

The full inventory (file:line, string, language) is posted on #871.

## This PR: trufi_core_home_screen

- `'Finding routes...'` → the existing `l10n.findingRoutes` (it was already used correctly elsewhere — plain regression).
- `'$totalWalkingMinutes min'` in the itinerary detail app bar → existing `durationMinutes` (renders **Min** in German).
- Routing-settings empty state was a full Spanish sentence → new `providerNoOptions` key (en/es/de), with the provider name as placeholder.
- Deep-link fallbacks `'Origin'`/`'Destination'` → new `placeOrigin`/`placeDestination` keys (en/es/de: Origin/Origen/Start, Destination/Destino/Ziel), resolved before the awaits so no BuildContext crosses an async gap.

`flutter gen-l10n` regenerated; package tests pass; analyzer clean (the one remaining info, `depend_on_referenced_packages`, is pre-existing).

## Remaining work (follow-up slices per package)

Per the inventory on #871: transport_list (+the es/en conditionals), saved_places, about, fares, ui (init screen + router), navigation (cubit messages + itinerary converter), maps, search_locations, and the OTP preference sheets in routing — the last three need l10n wiring from scratch.

PR #931, opened 2026-07-20
Fixes #852, #871: Resolve language inconsistencies

SidnyTech:
Fixes #852, #871 This pull request resolves language consistency issues across the application by improving localization support and replacing hardcoded user-facing strings with localized resources. It also updates application metadata and improves the localization infrastructure to make future translations easier to maintain.

### Localization improvements

- Added localized strings for newly introduced UI elements in English, Spanish, and German.
- Replaced hardcoded text with localized resources across the application.
- Added support for localized map engine descriptions and fare information using context-aware localization builders.

### Fare and map improvements

- Updated `FareInfo` and `FareCategory` models to support localized labels and descriptions.
- Updated fare widgets to display localized titles and category names.
- Localized map engine descriptions according to the active application language.

### About screen

- Added localized strings for the Contact section and its subtitle.
- Updated the About screen to use localized resources instead of hardcoded text.

## Result

- Language inconsistencies reported in **#852** and **#871** have been resolved.
- User-facing text is now consistently localized across supported languages.
- The localization infrastructure is more flexible for future features and translations.

### Preview

https://github.com/user-attachments/assets/50b73abc-d3f9-44e3-921f-8a84764b6382

## Notes

- While addressing #852 and #871, I verified all reported localization issues. Some strings initially identified as German translation issues were already correct German translations, so they were intentionally left unchanged. The changes in this PR focus only on actual inconsistencies and missing or hardcoded translations.

PR #940, opened 2026-07-22
fix(ui): localize init screen and router strings (#871)

guss1215:
Part of #871 — trufi_core_ui slice (init screen + router), 14 strings.

This one uncovered a real bug beyond the literals: `DefaultInitScreen` renders inside `AppInitializer`'s **inner** `MaterialApp`, which had no delegates, no locale and default `supportedLocales` — so even the existing `languageCode == 'es'` ternaries could never fire on the splash. The inner MaterialApp now gets `LocaleManager`'s locale plus `CoreLocalizations.localizationsDelegates`, and `DefaultInitScreen` resolves l10n defensively (English fallback if the delegate is absent, since it is publicly exported).

- 8 new keys (`initStepStarting/Initializing/LoadingMaps/LoadingRoutes/AlmostReady`, `errorUnableToStart`, `errorUnexpected`, `poweredByTrufi`); reused `appLoading`, `actionRetry`, `navAbout`, `titleError`, `errorPageNotFound`, `actionGoHome`.
- `poweredByTrufi` keeps the identical brand tagline in all three locales (apps can override).
- Two wording notes for review: the retry button reuses `actionRetry`, so en changes `'Try again'` → `'Retry'` (es already matched exactly); and for app locales outside {de,en,es} the splash now falls back to the first supported locale (`de`) instead of hardcoded English — if that matters we can add a `localeResolutionCallback` in a follow-up.

Verification: Translation Check replicated locally, `flutter analyze` clean (package has no test dir).

### DIFF
--- a/packages/trufi_core_ui/lib/l10n/core_de.arb
+++ b/packages/trufi_core_ui/lib/l10n/core_de.arb
@@ -19,5 +19,13 @@
   "actionGoHome": "Zur Startseite",
   "titleError": "Fehler",
   "unreadCount": "{count} ungelesen",
-  "markAllAsRead": "Alle als gelesen markieren"
+  "markAllAsRead": "Alle als gelesen markieren",
+  "initStepStarting": "Wird gestartet",
+  "initStepInitializing": "Wird initialisiert",
+  "initStepLoadingMaps": "Karten werden geladen",
+  "initStepLoadingRoutes": "Routen werden geladen",
+  "initStepAlmostReady": "Fast fertig",
+  "errorUnableToStart": "Start nicht möglich",
+  "errorUnexpected": "Ein unerwarteter Fehler ist aufgetreten",
+  "poweredByTrufi": "Powered by Trufi Association"
 }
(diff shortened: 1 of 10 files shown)


SamuelRioTz (approved):
**Verdict: ✅ Approved**

## Code review

- Fixes a real bug beyond l10n: the `AppInitializer` `MaterialApp` had no localization delegates or locale, so the init screen could never localize. This PR registers them and adds an English-fallback lookup — the parallel #931 touched the same file with a null-asserted lookup that would crash here instead.
- Preserves `widget.errorMessage ?? l10n.errorUnexpected` (real init errors still surface).
- Router error page (`errorPageNotFound`, `actionGoHome`) migrated to keys present in all three ARBs; generated files in sync.
- Overlaps only with #931 (core ARBs, `default_init_screen.dart`, `app_router.dart` — different keys); zero conflicts within the #936–943 series.
- Minor (non-blocking): unsupported locales now fall back to `supportedLocales.first` (currently German) on the splash — worth revisiting if more locales land.

Fresh install (cleared data) and cold starts in both languages, plus a bogus deep link for the router 404:

| Init screen (ES) | Init screen (DE) | Router 404 (DE) |
|---|---|---|
|  |  |  |

"Cargando mapas" ↔ "Karten werden geladen" — the splash localizes for the first time. 404 page: "Fehler / Seite nicht gefunden / Zur Startseite" (triggered via `https://trufi.example/app/pagina-inexistente`).

PR #941, opened 2026-07-22
fix(navigation): localize cubit and converter strings (#871)

guss1215:
Part of #871 — navigation slice, 13 strings across `NavigationCubit` and `ItineraryConverter` (neither has a `BuildContext`).

Best find of the batch: **9 of the 10 needed ARB keys already existed in en/es/de with real German translations — they were just never wired up.** Only `navRoute` is new.

Since the strings live in context-less code, they are now resolved at render time instead of being baked into state:

- `NavigationState` gains a `NavigationError` enum + optional `errorType`; the UI maps it to `NavigationLocalizations` at build time. The "open settings" action is now keyed off `errorType == permissionDeniedForever` instead of the fragile `errorMessage.contains('settings')` string match (legacy fallback kept for externally supplied messages).
- `NavigationInstruction` gains an optional `InstructionTextKey` (`youHaveArrived`, `finalDestination`) resolved at the widget layer — instructions re-localize on locale change and never persist English in state.
- `ItineraryConverter.toNavigationRoute` gains optional `originLabel`/`transferLabel`/`destinationLabel`/`routeLabel` params (English defaults preserve behavior for external callers; `NavigationScreen.showFromItinerary` passes localized values).

All API changes are additive; existing call sites compile unchanged. Added 7 pure-Dart unit tests (package previously had none).

Known follow-up (out of this slice's inventory): `'Exit at …'` / `'N stops'` / `'… remaining'` in `_buildTransitInstruction` are still hardcoded — their keys (`navExitAt`, `navOneStop`, `navStops`) also already exist unwired.

Verification: Translation Check replicated locally, `flutter analyze` clean, 7/7 tests pass.

### DIFF
--- a/packages/trufi_core_navigation/lib/l10n/navigation_de.arb
+++ b/packages/trufi_core_navigation/lib/l10n/navigation_de.arb
@@ -26,5 +26,6 @@
   "navLocationDisabled": "Standortdienste sind deaktiviert. Bitte aktivieren Sie diese.",
   "navOrigin": "Startpunkt",
   "navTransfer": "Umstieg",
-  "navDestination": "Ziel"
+  "navDestination": "Ziel",
+  "navRoute": "Route"
 }
--- a/packages/trufi_core_navigation/lib/l10n/navigation_en.arb
+++ b/packages/trufi_core_navigation/lib/l10n/navigation_en.arb
@@ -59,5 +59,7 @@
   "navTransfer": "Transfer",
   "@navTransfer": { "description": "Label for a transfer point between routes" },
   "navDestination": "Destination",
-  "@navDestination": { "description": "Label for the route destination point" }
+  "@navDestination": { "description": "Label for the route destination point" },
+  "navRoute": "Route",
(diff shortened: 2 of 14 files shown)


SamuelRioTz (approved):
**Verdict: ✅ Approved**

## Code review

- Architecturally the right pattern: the cubit emits enums (`NavigationError`, `InstructionTextKey`) and the widget layer resolves them to localized text — no l10n dependency leaks into business logic, consistent with the rest of the repo.
- 9 of 10 keys already existed with es/de/en translations; only one new key added, all three ARBs updated with generated files in sync.
- No file overlaps with any other open PR (including #931). Includes a unit test.

Built `apps/example` from this branch, ran real turn-by-turn navigation with simulated GPS in Cochabamba:

| Off-route banner (ES) | Exit dialog (ES) | German |
|---|---|---|
|  |  |  |

## Non-blocking follow-ups (already declared in PR description)

- "Exit at …" / "N stops remaining" instruction fragments remain in English (visible in screenshots) — tracked as follow-up.
- `copyWith` cannot clear `errorType` (minor).

PR #943, opened 2026-07-22
feat(maps,search_locations): add l10n infrastructure and localize strings (#871)

guss1215:
Part of #871 — the "screens with no l10n at all" slice: bootstraps localization infrastructure for `trufi_core_maps` and `trufi_core_search_locations` from scratch and localizes 25 strings (map settings, choose-on-map, engine names/descriptions, the whole search screen).

Infrastructure (mirrors `trufi_core_routing`):
- `l10n.yaml` + `maps_{en,es,de}.arb` / `search_locations_{en,es,de}.arb`, `flutter_localizations` + `generate: true` in both pubspecs, generated classes committed, `MapsLocalizations`/`SearchLocationsLocalizations` exported from the package barrels, tracked example `pubspec.lock`s updated.
- Delegates registered where they flow into apps: `HomeScreenTrufiScreen` (maps + search) and `TransportListTrufiScreen` (maps), plus the three affected example apps (maps, search_locations, poi_layers).

Pattern for context-less config/data classes: text fields become nullable (`ChooseOnMapConfiguration`, `SearchLocationBarConfiguration`, `LocationSearchScreenConfiguration`) and widgets resolve `null` via l10n; map engines get a `builder → displayName → l10n` fallback chain, which also fixes the Spanish-only `'Mapa vectorial claro/oscuro'`/`'Mapa completamente offline'` leaking into every locale. One documented skip: `'Vector (MapLibre)'` is a technical/brand label.

**Documented breaking edge**: widgets that resolve omitted text via these localizations now require the corresponding delegate when embedded in a standalone `MaterialApp` (TrufiApp-based apps are unaffected — delegates arrive via screen registration). Called out in the new README "Localization" section and in dartdoc on each affected widget; guarded by 6 new widget tests (en render, es render, and the delegate-less failure mode for both packages).

Verification: Translation Check replicated locally, `flutter analyze` clean in both packages, both examples, poi_layers example, home_screen and transport_list; maps 27/27, search_locations 63 passing with failures confined to the pre-existing live-network test files (byte-identical to main).

PR #944, opened 2026-08-03
fix: make N-language support actually work

SamuelRioTz:
Trufi is meant to ship in whatever languages a city translates it into — `gen_extra_l10n` (#919) exists precisely so a city can add Quechua, Aymara or Guaraní without touching core. Reviewing #931's `_localized(context, {es, de, en})` helper made me check whether that promise actually holds. It doesn't: adding a fourth language breaks the app in three separate ways, and none of them are caught by CI.

## What was broken

**1. The splash came up in German.** `AppInitializer` builds its own `MaterialApp` outside `_TrufiMaterialApp`, and it resolved against `CoreLocalizations.supportedLocales` (`[de, en, es]`) instead of the app's list. A city locale matched nothing, so Flutter fell back to that list's *first* entry — German — for every user of a Quechua app. It also ignored `extraLocalizationsDelegates`, so even a fully translated city language never reached the splash.

**2. Quechua/Aymara/Guaraní apps could not start.** `flutter_localizations` ships 82 languages and none of those three (verified against the SDK). With such a locale in `supportedLocales`, `GlobalMaterialLocalizations.delegate.isSupported()` returns false and the tree ends up with no `MaterialLocalizations` — every `Scaffold`, `Tooltip` and `showDialog` trips `debugCheckHasMaterialLocalizations`. The three languages the generator's own docs cite as the use case are exactly the ones that crashed.

**3. A partially translated language died on a null check.** Our generated getters are non-nullable (`nullable-getter: false`), so `CoreLocalizations.of(context)!` throws for any locale no delegate claims. Realistic case: a city translates most packages but not all, or adds the language before finishing the ARBs. Reproduced live — red screen, "Null check operator used on a null value".

## The fix

A fallback delegate layer in `trufi_core_ui`, registered **last** so real translations (including the app's own from `gen_extra_l10n`) always win:

- `fallbackFrameworkLocalizationsDelegates()` — serves Flutter's own strings from the app's default language for locales Flutter doesn't know.
- `fallbackLocalizationsDelegates([...])` — wraps core + screen delegates so an untranslated locale degrades to the default language instead of crashing.

Both `MaterialApp`s (splash and main) now use it, and the splash finally resolves against the app's own `supportedLocales` and extra delegates.

Plus, while in there:
- **Language names are app-extensible**: `TrufiLocaleConfig.languageNames` → `LocaleManager.displayName()`, so a city passes `{'qu': 'Runasimi'}` instead of the picker showing the bare code `QU`.
- **A stored language the app no longer ships is ignored**, instead of leaving the manager and the `MaterialApp` disagreeing.
- **The last two-language ternaries from #871 are gone** from `edit_place_dialog.dart` (`languageCode == 'es' ? … : …` — dead code behind a non-nullable getter, and it only ever knew Spanish and English). The ARB keys already existed; added `history`, the one label that was hardcoded and reachable.

## Verification

- New `n_language_support_test.dart` pins the promise: a Quechua app builds a `Scaffold` without asserting, **and** a German app still gets German framework strings (the fallback never shadows a real delegate).
- The example app now ships Quechua as a fourth language, so this stays exercised.
- Live on the Android emulator: the picker shows **Runasimi**, selecting it keeps the app fully usable (Settings, Home, route list) with Spanish as the fallback, zero exceptions in logcat.
- Full workspace suite run locally (CI does not run tests): no new failures; the four that fail are the pre-existing ones on `main` (two placeholder test files, the #859 scoring test, and network-dependent integration tests).

## Follow-ups, not in this PR

An audit for the same class of problem turned up more, worth separate PRs: `IRoutingProvider` needs `localizedName`/`localizedDescription` like `IMapEngine` already has (Trufi Planner's description is hardcoded Spanish inside core); `trufi_core_notifications` has no l10n infrastructure at all; GTFS mode names ("Bus", "Tram") are English constants; navigation instruction fragments have ARB keys that exist but are unused; and several `DateFormat('HH:mm')`/distance strings assume a fixed locale.


SamuelRioTz:
## Live verification on the Android emulator

The example app now declares Quechua as a fourth language (`supportedLocales: [es, en, de, qu]`, `languageNames: {'qu': 'Runasimi'}`) with **no Quechua ARBs anywhere** — the worst case this PR is meant to survive.

| Picker shows the endonym | Selected — no crash | Home works | Route list works |
|---|---|---|---|
|  |  |  |  |

- "Runasimi", not "QU" — the app-provided endonym.
- Selecting it leaves the app fully usable, falling back to Spanish (the app's default) for everything untranslated. Zero exceptions in logcat across Settings → Home → route list.
- Before this PR the same steps produced a red "Null check operator used on a null value" screen, and on a clean start the splash rendered in German.

A city that *does* run `gen_extra_l10n` gets its real Quechua strings instead of the fallback — the wrapper is registered last precisely so it never shadows a real delegate (pinned by the second test in `n_language_support_test.dart`, which asserts a German app still gets `"Zurück"`).

SamuelRioTz:
## Second pass: every user-visible fixed-language assumption is now closed

Commit `942d0d6f` extends this PR from "adding a language doesn't crash" to "adding a language actually localizes what you see". Live on the emulator in German — the leaks that survived the #936–#943 series:

| Trufi Planner + OTP descriptions | Map style descriptions | Line detail mode via l10n |
|---|---|---|
|  |  |  |

- "Läuft offline mit den in der App gebündelten GTFS-Daten" — the Spanish description that used to leak into every language now comes from `routing_*.arb` via the new `IRoutingProvider.localizedDescription` (same pattern `ITrufiMapEngine` already had).
- Navigation now uses the ARB keys that existed unused (`navExitAt`) plus a plural-aware `navStopsRemaining` (ICU plural, so languages with other plural rules work too).
- GTFS mode names, the notifications package (which had no l10n at all), share-route texts, and service-hours day names (which *crashed* for locales intl doesn't know) are all covered — details in the commit message.

Package suites re-run locally: all green except the four pre-existing failures on `main`.

SamuelRioTz:
**Review record** (self-merge with admin — GitHub blocks approving your own PR): full analysis run in-review, twice; the second pass caught and fixed a crash in the fallback layer itself and a missing delegate in the notifications example, both pinned by tests. ARBs aligned en/es/de, generated files in sync, analyze clean, all touched package suites green locally. Follow-ups in #945.

PR #946, opened 2026-08-05
fix(maps,transport_list): route detail camera never fits and snaps back on user movement

SamuelRioTz:
## Problem

Opening a route from the transport list shows the route geometry off-screen — the map never centers on it. Once the user pans or double-tap zooms, the map instantly jumps back, so gestures feel dead. Reported from on-the-ground testing of the Sana'a offline app ([trufi-sanaa#1](https://github.com/trufi-association/trufi-sanaa/issues/1)), but both defects are core-level and affect any app.

## Root causes (two independent defects)

1. **TrufiMap drops controlled cameras that arrive before the style loads.** `_moveCameraTo` only calls the native `moveCamera` when `_mapReady`; a camera set earlier only updates `_currentCamera` and is never replayed — and since `didUpdateWidget` compares against `_currentCamera`, it never retries either. With offline styles the route data always loads faster than the style, so the fitted camera always loses this race. (Online apps rarely hit it: the OTP fetch is slower than style load, which is why it went unnoticed.)

2. **`_RouteMapView` keeps a stale controlled camera.** `_onCameraChanged` never updated `_camera`, so every rebuild — e.g. the out-of-focus flip, which itself is triggered by the user's gesture — snapped the map back to the old fit. This is what made pans and double-tap zooms appear to "not work".

## Fix

- `trufi_map.dart`: remember a pending controlled camera and replay it in `onStyleLoadedCallback` (with `_suppressCameraCallback`, same as the normal path).
- `transport_detail_screen.dart`: track the user camera in `_onCameraChanged` so rebuilds pass the current position down instead of the stale fit. `_moveToLocation` / `_reFitCamera` keep working — they set `_camera` explicitly.

## Testing

- `flutter test` locally on `trufi_core_maps` and `trufi_core_transport_list`: **pass** (note: CI does not run tests).
- Verified on an emulator with a fully offline app (Sana'a): before/after below.

| Before | After (fitted) | After (pan persists) |
|---|---|---|
|  |  |  |

### DIFF
--- a/packages/screens/trufi_core_transport_list/lib/src/transport_detail_screen.dart
+++ b/packages/screens/trufi_core_transport_list/lib/src/transport_detail_screen.dart
@@ -1882,6 +1882,11 @@ class _RouteMapViewState extends State<_RouteMapView> {
   }
 
   void _onCameraChanged(TrufiCameraPosition cam) {
+    // Track the user's camera: this widget passes a controlled `camera` down,
+    // and leaving it stale makes every rebuild (e.g. the out-of-focus flip
+    // below) snap the map back to the old fit — pans and double-tap zooms
+    // appear to "not work" because they are instantly undone.
+    _camera = cam;
     final nowOutOfFocus = _fitCamera.isOutOfFocus(cam);
     if (nowOutOfFocus != _outOfFocus) {
       setState(() {
--- a/packages/trufi_core_maps/lib/src/presentation/map/trufi_map.dart
+++ b/packages/trufi_core_maps/lib/src/presentation/map/trufi_map.dart
@@ -670,7 +670,9 @@ class _TrufiMapState extends State<TrufiMap> implements TrufiMapDelegate {
                     1e-9 &&
                 (native.target.longitude - target.target.longitude).abs() <
                     1e-9 &&
-                (native.zoom - target.zoom).abs() < 1e-9;
+                (native.zoom - target.zoom).abs() < 1e-9 &&
+                (native.bearing - target.bearing).abs() < 1e-9 &&
+                (native.tilt - target.tilt).abs() < 1e-9;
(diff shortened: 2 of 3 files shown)


SamuelRioTz:
**Fresh-eyes audit** (independent reviewer instructed to attack these commits) came back: no blockers; three findings on the replay path, addressed in 5b3a27c2:

- the pending flag is now cleared unconditionally, so a stale camera can never replay on a later style load;
- the replay (and its callback suppression) is skipped when the native camera is already at the target — a no-movement `moveCamera` may emit no camera-idle on some platforms, which would have left `_suppressCameraCallback` armed and swallowed the next genuine gesture's callback;
- audited side effects on home_screen/navigation: the replay also fixes their previously-dropped pre-style camera moves; `_reFitCamera`/out-of-focus logic unaffected.

`flutter test` re-run on `trufi_core_maps` after the hardening: pass (27 tests).

SamuelRioTz:
**Second from-scratch review** (independent reviewer, no prior context) returned **minor changes**: the fix is correct — both root causes verified against main, no new medium/high defects — but the camera state machine had zero test coverage. Both conditions addressed in cbdd790b:

- the `alreadyThere` comparison is now a pure public helper, `camerasEffectivelyEqual`, covered by 10 unit tests (identical, fp-drift, per-component differences, bearing wrap cases). Bearing now compares by minimal angular difference, so the web (-180,180] vs native [0,360) conventions match instead of being a documented limitation;
- `_moveCameraTo` uses the same no-op skip, closing a pre-existing orphan suppression (imperative `moveCamera` with the current position could swallow the next gesture's callback).

Package tests: 37 passing. Emulator re-validated on this tip: route fit, pan persistence, double-tap, and the pan-during-style-load window.

PR #947, opened 2026-08-05
fix(utils): show the Arabic endonym in the language selector

SamuelRioTz:
## Problem

`LocaleManager.displayNameForCode` falls through to `code.toUpperCase()` for Arabic, so the onboarding and settings language selectors show **"AR"** instead of the endonym — while every other supported language shows its own name (English, Español, Deutsch…). First app affected: Sana'a (Arabic UI).

## Fix

One switch case: `'ar' → 'العربية'`.

## Testing

Manual, onboarding sheet on an emulator:

| Before | After |
|---|---|
|  |  |


SamuelRioTz:
Review question from Sam: instead of growing the endonym map by hand, why not ship **all the world's languages**? Done in the latest commit — `languageEndonyms` is now generated from the ISO 639-1 dataset (184 languages, `nativeName` first variant), keeping the curated capitalization for the languages Trufi already ships. Quechua (Runa Simi), Aymara, Guaraní, Amharic, Farsi and everything else resolve out of the box; unknown codes still fall back to the uppercased code, and per-app `languageNames` keeps priority. Covered by 5 unit tests (30 passing in the package).

PR #950, opened 2026-08-06
fix(home_screen): keep the viewport when switching map engines

### LINKED ISSUE #902: [Bug] [v5-Cochabamba] Map viewport resets (abrupt zoom-out) when changing map themes
**Bug description**

When switching between map styles (e.g., from OSM Bright to Fiord Color or Dark Matter), the application fails to preserve the user's current map state. The system ignores the active zoom level and coordinates, triggering an involuntary and abrupt "zoom-out" that resets the view to a macro level, losing the user's specific point of interest.

**Severity**
Priority: Medium (Affects navigation and UX)
Criticality: Medium (Major) - Users lose their current focus and must manually re-zoom and re-locate their destination after every theme change.

**Steps to Reproduce**
1. Open the Trufi Web App.
2. Set an Origin and Destination, and zoom in closely to a specific street or landmark.
3. Go to Map Settings.
4. Switch the style from OSM Bright to Fiord Color or Dark Matter.
5. Observe the sudden zoom-out behavior (Viewport Reset).
6. Gradually zoom out and compare the visibility of street names against the standard light mode (OSM Bright)

**Actual Behavior**
The map does not persist the  and  parameters during style updates. Every time a theme is selected, the viewport is reset to a default macro view.zoomcenter

**Expected Behavior**
The map must maintain the exact same zoom level and geographic coordinates (latitude/longitude) regardless of the visual theme selected..
   
**Attachments**

* Video:  

https://github.com/user-attachments/assets/79bf0ce1-7a34-4237-89f0-8cb0471bf6c8

(Demonstrating the abrupt zoom-out upon style change).

* Mobile

Image
Image
Image
Image

**Environment**
 - Device: Desktop PC & Mobile (Xiaomi Redmi Note 9 Pro).
 - OS: Windows 10 / Android 11.
 - Browser: Chrome / Edge.

SamuelRioTz:
Fixes #902.

## Problem

Changing the map style on the home screen resets the camera to the default city view. Each style in the map-type selector is a separate engine, and `setEngine` rebuilds the map with a brand-new widget — which started at the static `_initialCamera` (default center/zoom) instead of where the user actually was.

Reproduced on v5.20.0 (so the camera-lifecycle fixes from #946 don't cover this case — they act within a single map's life, not across engine swaps).

## Fix

One line in `_buildMap`: start the new engine's map at `_currentCamera` (the last camera reported by `onCameraChanged`, already tracked by the screen) and fall back to `_initialCamera` on first build.

Scope note: the transport detail screen is not affected (it drives the camera in controlled mode, which already carries across engine swaps), and `choose_on_map` can't reach an engine switch while open — so this is intentionally home-screen-only.

## Testing

- `flutter test` across all packages: green except `trufi_core_routing`, `trufi_core_search_locations`, `trufi_core_utils` — verified failing identically on clean `main` (preexisting: network-dependent integration tests, scoring test, placeholder files).
- Emulator before/after evidence in the review comment below.

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -1047,12 +1047,22 @@ class _HomeScreenState extends State<HomeScreen>
     }
   }
 
+  /// Moves the camera and records the destination in [_currentCamera].
+  /// Programmatic moves suppress the map's onCameraChanged callback, so
+  /// without this the recorded viewport goes stale and an engine switch
+  /// right after (e.g. my-location → change map style) snaps back to the
+  /// previous view (#902).
+  void _moveCameraTracked(TrufiCameraPosition position) {
+    _currentCamera = position;
+    _mapController?.moveCamera(position);
+  }
+
   Future<void> _onMyLocationPressed() async {
     // If already tracking, just center the map on current location
     if (_locationService.isTracking) {
       final location = _locationService.currentLocation;
       if (location != null) {
-        _mapController?.moveCamera(
+        _moveCameraTracked(
           TrufiCameraPosition(
             target: LatLng(location.latitude, location.longitude),
             zoom: 16,
@@ -1080,7 +1090,7 @@ class _HomeScreenState extends State<HomeScreen>
         // immediate feedback while the accurate fix is still being acquired.
         final lastKnown = await _locationService.getLastKnownLocation();
         if (lastKnown != null && mounted) {
-          _mapController?.moveCamera(
+          _moveCameraTracked(
             TrufiCameraPosition(
               target: LatLng(lastKnown.latitude, lastKnown.longitude),
               zoom: 16,
@@ -1095,7 +1105,7 @@ class _HomeScreenState extends State<HomeScreen>
         if (started && mounted) {
           final location = _locationService.currentLocation;
           if (location != null) {
-            _mapController?.moveCamera(
+            _moveCameraTracked(
               TrufiCameraPosition(
                 target: LatLng(location.latitude, location.longitude),
                 zoom: 16,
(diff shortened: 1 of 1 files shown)


SamuelRioTz:

- Root cause confirmed on v5.20.0: each map style is a separate engine, and `setEngine` rebuilds the map widget starting at the static default camera — the camera-lifecycle fixes from #946 act within one map's life and don't cover engine swaps.
- The fix reuses the screen's existing `_currentCamera` (updated on every `onCameraChanged`); `initialCamera` is only read when a map widget is created, so same-engine rebuilds are unaffected and the first build falls back to the default exactly as before.
- Caller audit: transport detail drives the camera in controlled mode (already carries across swaps); `choose_on_map` and navigation can't reach an engine switch while open. Home screen is the only affected caller.
- `flutter test`: green across packages except `trufi_core_routing` / `trufi_core_search_locations` / `trufi_core_utils`, which fail identically on clean `main` (preexisting, unrelated).

**Before (bug on v5.20.0)** — pan to Angostura, switch Liberty → Dark Matter, viewport resets to the default city view:

| Panned (Liberty) | After switch — RESET |
|---|---|
|  |  |

**After (this PR)** — same scenario, Dark Matter → Liberty, viewport stays:

| Panned (Dark Matter) | After switch — KEPT |
|---|---|
|  |  |

SamuelRioTz:
Fresh-eyes re-review of the first commit found a real gap (**applied in `afdee2a9`**): `_currentCamera` was only fed by user gestures — programmatic moves (the my-location button) suppress `onCameraChanged` by design, so tapping my-location and then switching map style still snapped back to the previous view. All programmatic moves now go through `_moveCameraTracked`, which records the destination before moving.

Verified on emulator — my-location (zoom 16) → switch Liberty → Dark Matter → viewport kept:

| My-location (Liberty) | After style switch — KEPT |
|---|---|
|  |  |

Re-review also confirmed: `initialCamera` is consumed once per widget creation (both shipped engines key their map by engine id, so a swap always recreates state); transport detail is immune (controlled camera); no home-screen flow relies on the old reset-to-default behavior. `flutter analyze` clean (one preexisting info) and `flutter test` green on the touched package.

PR #951, opened 2026-08-06
fix(maps): use hybrid composition on Android to stop platform-view crash

### LINKED ISSUE #948: [Bug][v5-Cochabamba] Application crashes after repeated map location selections
## 📝 Description
When selecting a location on the map (origin or destination), the map briefly flickers or refreshes unexpectedly. Repeating this action several times causes the application to become unstable and eventually crash.

This issue negatively impacts the route planning experience and may prevent users from successfully selecting locations on the map.

**Severity:** High – Can lead to application crash and loss of user session.  
**Priority:** High – Affects a core feature and degrades application stability.

---

## 📱 Test Environment
- **Device:** Redmi Note 11 Pro+
- **OS:** Android 13
- **Application:** Trufi App v5.1.0

---

## 🛠 Steps to Reproduce
1. Open Trufi App v5.1.0
2. Navigate to the route planner.
3. Tap the map to select an origin or destination.
4. Repeat the location selection process multiple times.
5. Observe the map behavior after each selection.

---

## 📎 Evidence

https://github.com/user-attachments/assets/46b41f1c-4378-49dc-ab8b-7a16b1754660

**Video:**  on this video we can see how the app crashes after selecting 3 times a location in the map.

---

## ❌ Actual Behavior
- The map flickers every time a location is selected.
- After several location selections, the application becomes unstable.
- Eventually, the application crashes and closes unexpectedly.

---

## ✅ Expected Behavior
- The selected location should be displayed smoothly without causing the map to flicker.
- Users should be able to select locations repeatedly without any visual glitches.
- The application should remain stable and should not crash regardless of how many times a user selects locations on the map.

### LINKED ISSUE #949: [Bug][v5-Cochabamba] Search causes map flicker and eventual app crash
📝 Description When using the search field in the Trufi app to find addresses or stops, the map briefly flickers or refreshes unexpectedly after selecting a search result. Repeating searches or selecting results multiple times makes the problem worse until the app becomes unstable and eventually closes.

Severity: High — Can lead to app crash. 
Priority: High — Affects a core feature and degrades user experience.

📱 Test environment

Device: Redmi Note 11 Pro+ (example)
OS: Android 16 (example)
App: Trufi App v5.1.0 (example)

🛠 Steps to reproduce

1- Open Trufi App
2- Go to the search field (search for addresses/stops).
3- Type a query and select a result.
4- Repeat the search/selection several times (try different results or reopen the search and select again).
5- Observe the map behavior after each selection and after multiple repetitions.

📎 Evidence

https://github.com/user-attachments/assets/bfd9ed1a-d7d5-4572-9711-d3c4e036610f

❌ Actual behavior

The map flickers / refreshes abruptly each time a search result is selected.
After several repeated selections, the app becomes unstable.
Eventually the app closes unexpectedly.

✅ Expected behavior

Selecting a search result should move the map smoothly to the chosen location without flicker.
Users should be able to repeat searches and selections without visual glitches.
The app should remain stable and must not close regardless of how many searches are performed.

SamuelRioTz:
Fixes #948, fixes #949.

## Problem

Repeatedly picking locations (on the map or from search) makes the map flicker and eventually crashes the app. Reproduced deterministically on emulator: **Trufi App v5.1.0 died in 9 iterations, the core example app (v5.20) in 4**, both with the identical fatal NPE on the main thread:

```
java.lang.NullPointerException: Attempt to invoke interface method
'int io.flutter.view.TextureRegistry$SurfaceProducer.getWidth()' on a null object reference
  at io.flutter.plugin.platform.SurfaceProducerPlatformViewRenderTarget.getWidth
  at io.flutter.plugin.platform.VirtualDisplayController.getRenderTargetWidth
  at io.flutter.plugin.platform.PlatformViewsController$1.lambda$resize$0
```

**Root cause:** maplibre_gl's docs claim hybrid composition is the default, but the actual default in the platform interface is `useHybridComposition = false` — every Trufi map runs in **Virtual Display** mode. That mode has a resize-after-dispose race: each choose-on-map / search flow creates and destroys a map platform view while the keyboard drives viewport resizes, and eventually a resize lands on a view whose SurfaceProducer was already released → NPE → crash. Memory stays flat across cycles (measured) — it's a lifecycle race, not a leak.

## Fix

Opt in to **hybrid composition** on Android (one guarded static set in TrufiMap's initState), which bypasses VirtualDisplayController entirely. The plugin's own docs note Virtual Display "may result in errors on Android 12" — this is that class of error.

Trade-off: hybrid composition can be slower on Android 9 and below; Virtual Display crashes on modern Android. The reported devices (Redmi, Android 13/16) are exactly the affected cohort.

## Testing

- Baseline (unfixed): crash in 4 iterations on the example app, 9 on Trufi App.
- Fixed build: **15 iterations clean**, empty crash buffer, flat memory, map renders and pans normally (screenshot in review comment).
- `flutter test`: trufi_core_maps 37/37; the 5 red packages (custom_material, planner, routing, search_locations, utils) fail identically on clean `main` (preexisting/flaky, verified today).

### DIFF
--- a/packages/trufi_core_maps/lib/src/presentation/map/trufi_map.dart
+++ b/packages/trufi_core_maps/lib/src/presentation/map/trufi_map.dart
@@ -2,7 +2,8 @@ import 'dart:async';
 import 'dart:math' as math;
 import 'dart:math' show Point;
 
-import 'package:flutter/foundation.dart' show kIsWeb;
+import 'package:flutter/foundation.dart'
+    show kIsWeb, defaultTargetPlatform, TargetPlatform;
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:maplibre_gl/maplibre_gl.dart' hide LatLngBounds;
@@ -140,6 +141,14 @@ class _TrufiMapState extends State<TrufiMap> implements TrufiMapDelegate {
   @override
   void initState() {
     super.initState();
+    // Render the native map with hybrid composition. Despite its docs, the
+    // plugin's actual default is Virtual Display, whose resize-after-dispose
+    // race kills the app with an NPE in VirtualDisplayController when map
+    // views are created and destroyed repeatedly — e.g. picking locations
+    // on the map or from search over and over (#948/#949).
+    if (!kIsWeb && defaultTargetPlatform == TargetPlatform.android) {
+      MapLibreMap.useHybridComposition = true;
+    }
     _currentCamera = widget.camera ?? widget.initialCamera;
     widget.controller?.attach(this);
   }


SamuelRioTz:

- Repro recipe (emulator, Pixel-class AVD): loop of *destination field → choose on map → pan → confirm*. Unfixed: fatal NPE at iteration 4 (example app) / 9 (Trufi App v5.1.0). Fixed: 15 iterations clean, crash buffer empty.
- Ruled out a memory leak first: TOTAL PSS flat (~570-590 MB) across 30 selection cycles on the unfixed build — the crash is purely the Virtual Display resize-after-dispose race.
- The change is a single guarded static (`!kIsWeb && Android`), set before the first map builds; iOS and web untouched.
- Rendering sanity with hybrid composition — style, labels, pins, pan all normal:



- Worth knowing: this also removes the per-selection flicker both issues describe — Virtual Display re-inits were the visible refresh.

SamuelRioTz:
One scoping note from the fresh re-review: this change eliminates the **crash** with certainty (the Virtual Display path is structurally gone — the flag makes the native MapView use `textureMode(true)`, so TLHC never falls back to VirtualDisplayController) and removes the keyboard-resize flash that Virtual Display re-inits caused. A *residual* brief flash when the choose-on-map view is created/destroyed can still occur (any freshly created TextureView flashes until its first frame) — if QA still sees a minor flicker on selection, that's view-churn territory and will be handled as a separate follow-up, not a reopen of the crash.

SamuelRioTz:
Upstream validation (web):

- The crash is exactly [flutter/flutter#154613](https://github.com/flutter/flutter/issues/154613) — `VirtualDisplayController.resize()` schedules a delayed runnable that runs after `dispose()` released the SurfaceProducer → NPE in `SurfaceProducerPlatformViewRenderTarget.getWidth()`. **Still open upstream** (labeled *needs repro info*), so there is no engine fix to wait for — avoiding the Virtual Display path is the right mitigation. Same crash class hit Mapbox's plugin with `textureView: false` ([mapbox-maps-flutter#887](https://github.com/mapbox/mapbox-maps-flutter/issues/887)).
- MapLibre's own [architecture docs](https://maplibre.org/flutter-maplibre-gl/concepts/architecture/) recommend exactly this flag: *"for improved performance and touch handling on Android 10 and later, set `MapLibreMap.useHybridComposition = true`"* — their example app enables it gated by SDK version. This PR sets it for all Android versions on the reasoning that a possible performance cost on Android ≤9 beats a deterministic crash; easy to gate by SDK later if field feedback asks for it.

Side note: since #154613 is tagged *needs repro info* and we have a deterministic 4-iteration repro recipe, it may be worth contributing it upstream.

PR #954, opened 2026-08-06
fix(routing): route lines no longer share the my-location blue

### LINKED ISSUE #930: [Bug][v5-Cochabamba] Route line and current location use the same color
### 📝 Description
The current location marker and the active route are displayed using the same blue color, making it difficult for users to distinguish between their position and the route being followed.

* **Severity:** Medium – Reduces visual clarity during navigation.
* **Priority:** Medium – UX issue that affects map readability.

### 📱 Test Environment
* **Device:** PC / Laptop
* **OS:** Windows 11
* **Browser:** Chrome Version 145.0.7632.109 (Official Build) (64-bit)

---

### 🛠 Steps to Reproduce
1. Open **Trufi App v5**.
2. Enable GPS.
3. Search for a route.
4. Display the route on the map.
5. Compare the colors used for the current location and the active route.

### 📎 Evidence
(Screenshot showing both the route and current location using the same blue color.)

### ❌ Actual Behavior
Both the user's current location and the active route are displayed using nearly identical colors.

### ✅ Expected Behavior
The route and the current location should use clearly distinguishable colors to improve map readability.

SamuelRioTz:
Fixes #930.

## Problem

The current-location dot and the active route line render in near-identical blues, so users can't tell their position from the route.

## Root cause (two layers deep)

The visible blue almost never comes from data: transit legs without a GTFS `route_color` get colors assigned **by index** from a default palette in `Itinerary._assignDefaultColors`, and the palette's first slot was `1976D2` — so the **first bus of nearly every itinerary** rendered in a blue that reads exactly like the my-location dot (`4285F4`). Measured on Cochabamba's bundled GTFS: **140 of 144 routes carry no route_color**, so this was the norm, not the exception.

## Fix

- `trufi_core_routing`: replace the palette's blue with deep purple (`673AB7`) — deliberately no blue in the default palette. Real GTFS `route_color` values are untouched (agency colors keep winning, per convention).
- `trufi_core_home_screen` (hardening): treat `FFFFFF` as unset when parsing leg colors (GTFS defines white as the *absence* of route_color — it would paint an invisible line on light maps) and fall back to the app's brand color instead of another blue if parsing fails.

The my-location dot keeps its Google-blue — that convention is universal and users expect it ([map UI patterns](https://mapuipatterns.com/blue-dot/)); per the [GTFS spec](https://gtfs.org/) route_color absent ⇒ FFFFFF, which is why the fallback layer exists.

## Testing

- Emulator before/after in the review comment: same 2-bus itinerary (Quillacollo → Plaza Libertad, offline data), first leg was dot-blue, now deep purple, with the blue location dot visible in the same frame.
- `trufi_core_routing` unit tests: +69 −10, failures **identical on the unchanged tree** (preexisting HTTP-mock tests); no test asserts palette colors. `trufi_core_home_screen`: 7/7. Analyze clean in both (2 preexisting infos in routing).

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -2751,12 +2751,9 @@ TrufiLocation _searchLocationToTrufiLocation(SearchLocation location) {
   );
 }
 
-Color _parseColor(String hexColor, {Color fallback = const Color(0xFF1976D2)}) {
+Color _parseColor(String hexColor, {required Color fallback}) {
   try {
     final hex = hexColor.replaceFirst('#', '');
-    // GTFS treats an absent route_color as white (FFFFFF) — that's "no
-    // color", not a paintable line color, so fall back too.
-    if (hex.toUpperCase() == 'FFFFFF') return fallback;
     return Color(int.parse('FF$hex', radix: 16));
   } catch (_) {
     return fallback;
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -826,8 +826,13 @@ class _HomeScreenState extends State<HomeScreen>
       final bool useDots;
 
       if (leg.transitLeg) {
-        // Transit legs use route color
-        color = _parseColor(leg.routeColor);
+        // Transit legs use the GTFS route color; without one, fall back to
+        // the app's brand color instead of a blue that reads exactly like
+        // the my-location dot (#930).
+        color = _parseColor(
+          leg.routeColor,
+          fallback: Theme.of(context).colorScheme.primary,
+        );
         lineWidth = 6;
         useDots = false;
       } else if (leg.transportMode == routing.TransportMode.bicycle) {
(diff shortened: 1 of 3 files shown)


SamuelRioTz:

- The diagnosis went two layers deep: the first attempt (theme-color fallback in the home screen) turned out to be **downstream dead code** for the reported case — `Itinerary._assignDefaultColors` pre-fills every colorless transit leg from the palette, so `leg.routeColor` is never empty by the time the map paints it. The palette itself was the bug; the home-screen change stays as hardening for the `FFFFFF`/parse-failure edge cases.
- Verified against the data: 140/144 routes in Cochabamba's bundled GTFS have no `route_color` → palette slot 0 (`1976D2` blue) was on screen in virtually every plan.
- Baselines: the 10 failing routing unit tests fail identically on the unchanged tree (preexisting HTTP-mock tests); no test asserts palette colors.

**Before** — first transit leg takes palette blue, same hue as the location dot · **After** — deep purple, dot clearly distinct (same frame):

| Before | After |
|---|---|
|  |  |

SamuelRioTz:
Fresh-eyes re-review (CAMBIOS MENORES) — all three findings applied in `c6cdc67b`:

- **F1**: the `FFFFFF` special-case moved from the home screen to `Itinerary._assignDefaultColors` (model layer). A leg with a literal `route_color=FFFFFF` previously skipped the palette and painted **white-invisible** in the itinerary chip, the detail timeline and the transport-list map while the home map showed the theme color — now every screen agrees, and two FFFFFF legs get distinct palette colors.
- **F2**: `673AB7` measured 2.51:1 against the dark map style — replaced with the lighter `7E57C2` (verified on the dark style, screenshot below).
- **F3**: 4 unit tests now pin the guarantees: palette slot 0 is never dot-blue, FFFFFF is treated as unset, colorless legs get distinct colors, real GTFS colors are preserved. All green.

Also noted by the re-review, left as follow-ups: the same dot-blue collision exists in the navigation layer (`navigation_layer.dart` falls back to `Colors.blue` — separate issue territory), and previously persisted plans keep their baked-in blue until the next search (self-corrects).

PR #956, opened 2026-08-09
fix(saved_places): reject duplicate places with a friendly warning

### LINKED ISSUE #898: [Bug] [v5] Application allows saving duplicate names and locations in "Your places" (Cross-platform)
**Description**

The application allows users to save multiple entries in the "Your places" section that share both the exact same name and the same geographic location. The system does not trigger any warning or prevent the creation of these redundant entries.
This issue occurs on both the Web/Desktop version and the Mobile application.

- **Criticality:** Medium (Partial functionality issue / annoying but doesn't prevent usage).
- **Priority:** P3 (Relevant Improvement / Third Highest).

**Test environment**

**Desktop:**
- Device: Desktop PC.
- OS: Windows 11.
- Browser: Microsoft Edge v145.

**Mobile:**

- Device: Honor Magic7 Lite.
- OS: Android 16 (MagicOS 10.0).
- App Version: v5.0.0 (Cochabamba).

**Steps to Reproduce**

1. Open the app (Web or Mobile) and navigate to the side menu.
2. Select "Your places" (Tus lugares).
3. Tap the "+" icon and confirm a specific location on the map.
4. Enter a name for this location and save it.
5. Repeat steps 3 and 4 using the exact same name and selecting the exact same location on the map.
6. Observe the "Your places" list.

**Evidence**

**Mobile:**

https://github.com/user-attachments/assets/3cd23f68-2243-4fe4-b988-b651d57a0ed3

**Desktop:**

https://github.com/user-attachments/assets/d78e5512-0a66-4171-a375-20522985a490

**Actual Behavior**

The application saves the duplicate entry (same name and location) without any warning, resulting in a redundant list.

**Expected Behavior**

The system should detect that an entry with the same name and/or coordinates already exists and show a warning message to prevent duplication.

SamuelRioTz:
Fixes #898.

## Problem

"Your Places" silently accepts multiple entries with the exact same name and the same coordinates — no warning, no prevention (reproduced on v5.20: three identical "Prueba" favorites).

## Fix

- New `SavedPlacesCubit.isDuplicatePlace(place, {excludeId})`: same name (trimmed, case-insensitive) at the same coordinates (1e-6 ≈ sub-meter epsilon — the stored values of "identical" map-picker selections differ by ~1e-14) against home, work and favorites. `excludeId` lets the edit flow avoid self-collision.
- Both add/edit flows now check it and show a localized snackbar (`placeAlreadySaved` in en/es/de) instead of saving.
- 6 unit tests with an in-memory repository.

## Honest notes

- The dialog wiring is **duplicated** between `saved_places_trufi_screen.dart` (the screen apps actually register) and `widgets/saved_places_screen.dart` — the first fix round only patched the latter and the emulator caught it (a third duplicate went straight through). Both are patched now; unifying them is tech debt for a follow-up.
- Same name at a *different* location (or different name at the same spot) is still allowed on purpose — only exact duplicates are rejected.

## Testing

- `flutter test` on the package: 16/16 (10 existing + 6 new). Analyze clean.
- Emulator before/after: duplicate save silently accepted before; now the snackbar shows and no entry is created (screenshots in review comment).

### DIFF
--- a/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
+++ b/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
@@ -34,5 +34,6 @@
   "change": "Ändern",
   "errorLoadingPlaces": "Fehler beim Laden der Orte",
   "retry": "Erneut versuchen",
-  "history": "Verlauf"
+  "history": "Verlauf",
+  "placeAlreadySaved": "„{placeName}\" ist an diesem Ort bereits gespeichert"
 }
--- a/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
+++ b/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_de.arb
@@ -35,5 +35,5 @@
   "errorLoadingPlaces": "Fehler beim Laden der Orte",
   "retry": "Erneut versuchen",
   "history": "Verlauf",
-  "placeAlreadySaved": "„{placeName}\" ist an diesem Ort bereits gespeichert"
+  "placeAlreadySaved": "„{placeName}“ ist an diesem Ort bereits gespeichert"
 }
--- a/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_en.arb
+++ b/packages/screens/trufi_core_saved_places/lib/l10n/saved_places_en.arb
@@ -46,5 +46,14 @@
   "history": "History",
   "@history": {
     "description": "Label for saved places coming from search history"
+  },
+  "placeAlreadySaved": "\"{placeName}\" is already saved at that location",
+  "@placeAlreadySaved": {
+    "description": "Shown when saving a place that duplicates an existing one (same name and location).",
+    "placeholders": {
+      "placeName": {
+        "type": "String"
+      }
+    }
   }
 }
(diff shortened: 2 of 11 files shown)


SamuelRioTz:

- Repro on v5.20: two (then three) identical "Prueba" favorites saved without any warning — the stored coordinates of repeated map-picker selections differ only at the ~1e-14 level, which informed the 1e-6 epsilon.
- First fix round patched only `widgets/saved_places_screen.dart`; the emulator retest caught that the app-registered screen (`saved_places_trufi_screen.dart`) has its **own copy** of the same dialog wiring and still saved a third duplicate. Both flows are guarded now; unifying the duplicated wiring is noted as tech debt.
- UX reference: Google Maps' save flow is a toggle per list, so exact duplicates are impossible by construction there — a form-based flow like ours needs the explicit check ([Google Maps Help](https://support.google.com/maps/answer/3184808)).

| Before — duplicates accepted silently | After — snackbar, nothing saved |
|---|---|
|  |  |

SamuelRioTz:
Fresh-eyes re-review (CAMBIOS MENORES) — findings applied in `f9e5b25c`:

- **Main finding**: users who already have duplicates from before this guard could no longer edit either copy — editing one (icon, type) collided with its pre-existing twin and the edit was silently discarded. Fixed: an edit that keeps the same name and coordinates skips the check (it can't create a *new* duplicate). Verified on emulator using the duplicates created during the repro: changing only the icon on one of three identical "Prueba" entries now saves normally (star icon below), still three entries.
- German quotes corrected („…“), epsilon-edge tests (1e-7 in / 1e-5 out) and identity tests added — 18/18 green.
- Noted by the review, out of scope: the guard is UI-level (the cubit itself still accepts duplicates — no other callers exist in the monorepo today) and names aren't Unicode-normalized (NFC vs NFD accents). Both fit the dialog-wiring unification follow-up.

PR #957, opened 2026-08-09
fix(transport_list): rank exact short-name matches first in route search

### LINKED ISSUE #844: [Bug] [v5-Cochabamba] Route "R" not found in search results
**Bug description**

The route "R" cannot be found when searching within the application. When users attempt to search for the line using the search bar, no results are returned.
This issue affects route discoverability and prevents users from accessing route details.

**Severity**

- *Criticality*: Medium
- *Priority*: High

**Test Environment**

- Device: Honor X6b
- OS: Android 14
- App Version: v5.0.2 (Cochabamba)

**Steps to Reproduce**

1. Open the Trufi App.
2. Navigate to the route search section.
3. Enter "R" in the search bar.
4. Observe the search results.

**Current Result**

The system returns no results for route "R".

**Expected result**

Route "R" should appear in the search results and allow users to access its route details and map visualization.
   
**Attachments**

Image

SamuelRioTz:
Fixes #844.

## Problem

Searching "R" in the routes list returns every route that *contains* the letter anywhere ("MiniBus 1 **Verde**", "**Trufi** 01"...) with no ranking — the actual bus **R** is buried hundreds of rows deep. Reported on Cochabamba (657 routes) and confirmed by the reporter after an earlier close.

## Fix

The base provider's filter now sorts matches by quality before publishing them:
1. exact short-name match (case-insensitive, trimmed)
2. short-name prefix
3. short-name contains
4. matches only in long name / headsign / agency

Original list order is preserved within each rank (index-decorated sort — `List.sort` isn't stable). The filter itself is unchanged: same fields, same substring semantics, so no previously-findable route disappears. One fix covers both GTFS and OTP providers (shared base class).

## Testing

- 4 new unit tests on the ranking (exact-first for single-letter queries, case-insensitivity, exclusion unchanged, clear-query restores original order); package suite 16/16 green, analyze clean.
- Emulator before/after: "R" used to surface "MiniBus 1 (Verde)" first; now "Línea Roja (R)" and "MicroBus R" lead the list (screenshots in review comment).

### DIFF
--- a/packages/screens/trufi_core_transport_list/lib/src/transport_list_data_provider.dart
+++ b/packages/screens/trufi_core_transport_list/lib/src/transport_list_data_provider.dart
@@ -154,12 +154,14 @@ abstract class TransportListDataProvider extends ChangeNotifier {
         return 3;
       }
 
-      final decorated = matches.asMap().entries.toList()
-        ..sort((a, b) {
-          final byRank = rank(a.value).compareTo(rank(b.value));
-          return byRank != 0 ? byRank : a.key.compareTo(b.key);
+      final decorated = [
+        for (var i = 0; i < matches.length; i++)
+          (rank: rank(matches[i]), index: i, route: matches[i]),
+      ]..sort((a, b) {
+          final byRank = a.rank.compareTo(b.rank);
+          return byRank != 0 ? byRank : a.index.compareTo(b.index);
         });
-      result = decorated.map((e) => e.value);
+      result = decorated.map((e) => e.route);
     }
 
     _state = _state.copyWith(filteredRoutes: result.toList());
--- a/packages/screens/trufi_core_transport_list/lib/src/transport_list_data_provider.dart
+++ b/packages/screens/trufi_core_transport_list/lib/src/transport_list_data_provider.dart
@@ -134,12 +134,32 @@ abstract class TransportListDataProvider extends ChangeNotifier {
 
     if (_query.isNotEmpty) {
       final lowerQuery = _query.toLowerCase();
-      result = result.where((route) {
+      final matches = result.where((route) {
         final searchText =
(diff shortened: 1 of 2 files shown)


SamuelRioTz:

- Root cause: plain substring `contains` over the concatenated fields with no ordering — with 657 routes, a single-letter query matches nearly everything ("Ve**r**de", "T**r**ufi"...), so the count barely drops and the exact route drowns.
- The sort is decorated with the original index because `List.sort` is not stable — ties keep the incoming order, so agency grouping/original ordering is untouched within ranks.
- Scope note: this is the ROUTES-list search. The issue's title ("route R not found") matched what the reporter's second comment clarified: it's a ranking problem, not a matching one.

| Before — "R" buried under letter-matches | After — exact "R" routes lead |
|---|---|
|  |  |

SamuelRioTz:
Fresh-eyes re-review (CAMBIOS MENORES) — applied in `625a94c1`:

- **Main finding worth reading**: the original stability test (5 routes) could never catch the index-decoration being removed, because Dart's `List.sort` uses insertion sort — de-facto stable — below ~32 elements; only at production sizes (657 routes) would ties visibly shuffle on every keystroke. A 120-route all-ties fixture now pins the guarantee.
- Rank is precomputed once per match (records decoration) instead of ~2·n·log n times inside the comparator.
- Added the missing edges: null-shortName routes matching via other fields, and inner-space queries.
- Confirmed by the reviewer (no changes needed): during search the list renders FLAT — no agency grouping to override the ranking; the count chip is order-agnostic; agencyFilter composes correctly. Design note recorded: an exact longName match deliberately ranks below shortName matches (shortName is the route's identity — the exact ask of #844).

19/19 tests, analyze clean.

Issue #830, opened 2026-02-23
[Bug][v5-Cochabamba] Locations selected in the origin and destination are cleared after refreshing the page/app


jheimc:
### 📝 Description
When performing a route search by manually selecting locations on the map, refreshing or reloading the page (F5 or manually refreshing the page) clears the selected locations (GPS coordinates) from the search fields. 
However, the route remains active and the user can continue using the application without needing to re-enter the route. This issue only affects the visibility of the coordinates in the search bar and may cause confusion, as the displayed fields no longer reflect the current route.

* **Criticality:** Low - State Persistence Issue

### 📱 Test Environment
* **Device:** Mobile: Redmi Note 11 Pro+ 
* **Device:** PC / Laptop 
* **OS:** Windows 11
* **Browser:** Chrome Version 145.0.7632.109 (Official Build) (64-bit)

### 🛠 Steps to Reproduce
1.	Open the browser and go to the URL: `https://planner.trufi.app/`
2.	Select any origin and destination of a known route (using the Choose in the Map option)
or open the following link: https://planner.trufi.app/?from_lat=-17.396030&from_lng=-66.153571&from_name=Selected+Location&to_lat=-17.379943&to_lng=-66.160364&to_name=Selected+Location&time_mode=leaveNow&route=0
3.	perform a page refresh (if using the mobile app perform an Swipe Down action).
4.	Observe the Origin and Destination loses the GPS coordinates after the refresh.

### Evidence

<img width="1917" height="865" alt="Image" src="https://github.com/user-attachments/assets/07dac11c-6b65-4f83-8e0e-4d75a16ce40a" />

<img width="1918" height="861" alt="Image" src="https://github.com/user-attachments/assets/f158ba62-bad1-47e1-aba2-2ab1698860c3" />

### ❌ Actual Behavior
After refreshing or reloading the page, the selected origin and destination (GPS coordinates) are cleared from the search fields.

### ✅ Expected Behavior
After refreshing or reloading the page, the previously selected origin and destination (GPS coordinates) should remain visible in the search fields. The search bar should accurately reflect the currently active route, ensuring consistency between the displayed information and the application state.

SamuelRioTz:
Retested on current core with a headless-Chrome harness (web build served locally, driven over the DevTools protocol) — **this no longer reproduces**. After setting origin and destination and pressing reload, everything comes back:

- both search fields ("Quillacollo" / "Plaza Cristobal Colón"),
[...]

PR #958, opened 2026-08-09
fix(example): point the web planner at the gateway's /api prefix

SamuelRioTz:
## Problem

The example app's **web** build can never plan a route with Trufi Planner: every request fails with

```
FormatException: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

## Root cause

The planner API lives behind the gateway's `/api` prefix, but the example configures `serverUrl: 'https://planner.trufi.app'`. Every call therefore lands on the web app and comes back as the HTML shell. Verified against production:

| Request | Result |
|---|---|
| `POST /api/plan` | `{"success":true,"paths":[…]}` — real itineraries |
| `POST /plan` | `<!DOCTYPE html>…` — the app shell |
| `GET /health` and `/api/health` | `{"status":"healthy"}` (which is why the provider looked reachable) |

The Cochabamba app already gets this right (`serverUrl: '$_baseUrl/api'` in trufi-app), so this is example-only — but the example is what city apps are copied from, so a wrong URL there propagates.

## Fix

One line plus a comment explaining the prefix.

## Testing

Headless Chrome against the built web app: before the change the planner returned the HTML error above; after it, the same origin/destination returns **5 itineraries**, and reload restores the full state (fields, plan, selected route). Screenshots in the review comment.

### DIFF
--- a/apps/example/lib/main.dart
+++ b/apps/example/lib/main.dart
@@ -50,7 +50,10 @@ final List<IRoutingProvider> _routingEngines = [
   if (kIsWeb)
     TrufiPlannerProvider(
       config: const TrufiPlannerConfig.remote(
-        serverUrl: 'https://planner.trufi.app',
+        // The planner API lives behind the gateway's /api prefix; without
+        // it every request lands on the web app and comes back as HTML
+        // ("FormatException: Unexpected token '<'").
+        serverUrl: 'https://planner.trufi.app/api',
       ),
     ),
   // Online routing via OTP (dev servers)


SamuelRioTz:

Found while retesting #830 on a headless Chrome harness (the web app served locally, driven over the DevTools protocol).

| Before — planner unreachable, HTML parsed as JSON | After — 5 itineraries | After reload — full state restored |
|---|---|---|
|  |  |  |

Note on scope: I also probed a theoretical window in the URL-restore path (origin and destination are applied in two steps, so a listener firing in between could write a partial URL). I could **not** reproduce it in 6 runs across two failure scenarios, so I dropped that speculative guard and kept only the change with evidence behind it.

Issue #953, opened 2026-08-06
Operator QR Code functionality fails (No redirection, copy error, plain text link)


IsmaDavidAA:
Description:
The QR code sharing feature for transport operators in the mobile application is currently experiencing multiple failures. Specifically, the generated QR code does not redirect to any destination, the "Copy" functionality triggers an error, and the shared link output lacks URL formatting, rendering it as plain text.

Steps to Reproduce:
Open the Trufi app and navigate to an operator's detail view (e.g., "Asociación de autotransporte Uspa Uspa").
Open the QR code sharing modal ("Scan to open this operator's routes").
Issue 1: Scan the displayed QR code with a camera or scanner app. (Observe that it does not redirect anywhere).
Issue 2: Tap the "Copy" button located under the QR code. (Observe the system toast message: "Couldn't copy the QR").
Issue 3: Tap the "Share" button to send the link via a messaging app. 

Expected Behavior:
Scanning: The QR code should successfully redirect the user to the specific operator's routes within the app via a deep link, or to a fallback web page if the app is not installed.

Copying: The "Copy" button should successfully copy the QR code image [()](trufiapp://routes?operator=Asociaci%C3%B3n+de+autotransporte+Uspa+Uspa) to the device's clipboard without throwing an error.

Sharing: The shared text should be formatted as a valid, clickable hyperlink.

Actual Behavior:
The QR code fails to redirect the user.
A toast message reading "Couldn't copy the QR" is displayed when attempting to copy.
The shared link is output as unclickable plain text.

Environment:
Application: Trufi App (Mobile)
OS: Android

Attachments:

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/5434d24e-dc68-46b1-b622-b4049b2bfed9" />

SamuelRioTz:
Fixed across both repos — trufi-core#959 and trufi-association/trufi-app#40, both merged.

**1. The QR opened nothing.** The app built the routes screen without `shareBaseUrl`, so the code encoded the custom scheme `trufiapp://routes?operator=…` — a camera app cannot resolve that. (It could not have worked inside the app either: it produced host `routes` while the manifest registers `route`.) It now encodes `https://planner.trufi.app/routes?operator=…`, and core refuses to build a QR at all from anything that isn't an http(s) base — a hidden button beats a printed code that opens nothing.

[...]

PR #959, opened 2026-08-09
fix(transport_list): only encode scannable links in the operator QR

SamuelRioTz:
Core half of #953 (the host-app half is trufi-association/trufi-app#40).

## Problem

The QR sheet fell back to the app's custom scheme (`trufiapp://routes?operator=…`) whenever `shareBaseUrl` wasn't configured. **A camera app cannot resolve a custom scheme**, so operators printed codes that opened nothing — the worst possible failure for something that ends up glued inside a vehicle.

Two smaller defects in the same sheet:
- **Share sent plain text.** `ShareParams(text: link)` leaves the receiver with an untappable string in most messengers.
- **The copy error was swallowed.** `debugPrint('… failed: \$e')` had the `$` escaped, so the log printed the literal `$e` and the real exception never appeared. Unescaping it immediately revealed the actual cause of the copy failure (a FileProvider root missing in the host app — fixed in the app PR).

## Fix

- `_operatorLink` builds a code **only** from the https share URL; without it the QR affordances stay hidden (a hidden button beats a printed code that does nothing). `_canShareOperatorQr` follows the same rule.
- Share as `ShareParams(uri: …)` when the link is http(s) so messengers linkify it; plain text stays as the fallback.
- Unescape the diagnostic so copy failures are debuggable.

## Testing

Package suite 19/19 green, analyze clean. End-to-end on emulator with the app PR applied: the sheet shows the https link, and **Copy QR** completes with Android's clipboard preview showing the image (screenshots in the app PR).

Behaviour note for embedders: an app that configured only `deepLinkScheme` and no `shareBaseUrl` will now see the QR buttons disappear instead of producing an unscannable code. That is deliberate.

### DIFF
--- a/packages/screens/trufi_core_transport_list/README.md
+++ b/packages/screens/trufi_core_transport_list/README.md
@@ -1,0 +1,54 @@
+# trufi_core_transport_list
+
+Routes list, route detail, and the operator QR sheet.
+
+## Host-app requirements
+
+### Operator QR codes
+
+Pass `shareBaseUrl` to `TransportListTrufiScreen` with the app's **https**
+base (e.g. `https://planner.trufi.app`). The QR encodes
+`<base>/routes?operator=<agency>`.
+
+A QR is read by a camera app, which resolves http(s) and nothing else — a
+custom scheme such as `trufiapp://` produces a code that opens nothing, so
+without a usable https base the QR affordances stay hidden by design.
+
+For the scanned link to open the **app** (rather than the website) the host
+app must register it as an App Link / Universal Link: an intent filter for
+that host with `android:autoVerify="true"`, plus a matching
+`assetlinks.json` (Android) and `apple-app-site-association` (iOS) served
+from the domain. On Android the verification uses the **signing key the
+store re-signs with**, so App Links only resolve on store-installed builds
+— a locally built APK opens the browser instead. Without any of this the
+link still works: it falls back to the web app.
+
+### Copying the QR image
+
+"Copy QR" hands a rendered PNG to the clipboard through the `pasteboard`
+plugin, which needs a `FileProvider` in the **host app's** manifest:
+
+```xml
+<provider
+    android:name="androidx.core.content.FileProvider"
+    android:authorities="${applicationId}.provider"
+    android:exported="false"
+    android:grantUriPermissions="true">
+    <meta-data
+        android:name="android.support.FILE_PROVIDER_PATHS"
+        android:resource="@xml/provider_paths" />
+</provider>
+```
+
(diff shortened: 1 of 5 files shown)


SamuelRioTz:

Diagnosis went in two steps because the code hid its own error: the copy failure logged a literal `\$e`, so after unescaping it the real exception appeared and named the cause —

```
PlatformException(Error, Failed to write image, Failed to find configured root
that contains /data/data/app.trufi.navigator/cache/….png)
```

The pasteboard plugin's README only documents `<external-path>`, but the PNG is written to the app's **internal cache**, so `<cache-path>` is the root that matters — worth knowing for every city app that adds this feature.

| QR now encodes the https link | "Copy QR" completes (clipboard preview) |
|---|---|
|  |  |

Tested on an emulator with the core branch wired into the app through a temporary path override (reverted before committing).

SamuelRioTz:
Fresh-eyes re-review (CAMBIOS MENORES) — code findings applied:

- **The gate now checks scannability, not just null.** `shareBaseUrl: "planner.trufi.app"` (no scheme) used to build `///routes?operator=X` and encode garbage; a `trufiapp://` base brought back the original bug. The rule lives in a pure `operatorQrLink()` helper with 6 unit tests (25 total in the package, all green).
- **Docs corrected**: the public dartdoc still advertised the removed custom-scheme fallback, and the share comment claimed a URI share differs on Android — it does not (`Share.kt` builds the same intent); what makes the message tappable is the https link. Both fixed.
- **New package README** documenting the two host-app requirements nobody had written down: App Links for the scan to open the app, and the FileProvider/`cache-path` for the copy. The app PR now declares only `cache-path` (the other roots were dead config), re-verified on emulator.
- Test fixture moved off `trufiapp://`, which the product can no longer emit.

**Two operational blockers the review surfaced — code alone will not close #953:**

1. **The web fallback is stale.** `planner.trufi.app` serves **5.0.3+75** (bundle from May), which predates the `operator` filter — someone without the app who scans the QR lands on the routes list **unfiltered**. The web build has to be redeployed from `main` alongside this fix.
2. **App Links only verify for store-signed builds.** The live `assetlinks.json` lists two SHA-256 fingerprints, and neither is our upload key or a debug key — expected, since Play re-signs. Consequence for QA: a locally built APK opens the **browser**, not the app. This must be validated from an internal-testing build, or it will be reported as "still broken".

Also worth knowing: apps without an https base (trufi-sanaa, the examples) lose the QR button with this change. Deliberate — the codes they emitted could not be scanned — but it is a silent feature removal, and the package has no CHANGELOG to record it.

Issue #820, opened 2026-02-22
[Bug] [High] [v5-Cochabamba] Destination field displays GPS coordinates instead of street names


MelCampero11:
**Describe the bug**
When a point is selected on the map, the destination field does not display the street name or area. Instead, it only shows raw GPS coordinates (e.g., -17.4444...). Consequently, when trying to find a route, the app returns "0 routes found" or displays irrelevant communities.

- Priority: High
- Criticality: High

**Steps to Reproduce**
1. Open the Trufi Web App.
2. Click on the "Destination" input and select "Choose on map".
3. Move the marker to a specific area (e.g., a known street in Cochabamba).
4. Confirm the point and observe the text displayed in the destination field.
5. Attempt to search for a route to that location.

**Actual Behavior**
The destination field only displays coordinate numbers. The route search either fails or provides irrelevant results because it cannot translate the coordinates into a valid destination name.

**Expected Behavior**
The field should automatically populate with the street name, neighborhood, or a recognizable landmark (Reverse Geocoding) so the user can verify the location and find valid routes.

**Screenshots**

<img width="1794" height="729" alt="Image" src="https://github.com/user-attachments/assets/01a1c4ed-eb38-4b1c-b1bc-988bafae02c9" />

**Test Environment**

- Device: Desktop PC / Laptop
- OS: Windows 10
- Browser: Chrome / Edge

IsmaDavidAA:
@MelCampero11 please could you share the request that you used to report this bug?

SamuelRioTz:
Closing — fixed. Picking a point on the map now resolves to a **street name**, not raw coordinates:

<img src="https://raw.githubusercontent.com/trufi-association/trufi-core/1eda29e8ca3f5f4aa2379bb421d96b09a75136dc/core-issues/820/01-campo-tras-elegir-en-mapa.png" width="420">

The destination field reads *"Pasaje A, Cochabamba, Cochabamba…"* and the shared URL carries `to_name=Pasaje+A`. What fixed it is the reverse-geocoding step added in the v5.19/v5.20 rework: the picked point is resolved (behind a progress indicator, with a guard so a slow answer can't overwrite a newer pick) before it reaches the field.

Note for whoever validates on the live site: **planner.trufi.app currently serves 5.0.3+75**, a build from May that predates this — so the fix will only be visible there after the next web deploy. Verified on a build from current `main` (5.1.0+80).

PR #960, opened 2026-08-09
fix(maps): tapping the map picks that point in choose-on-map

### LINKED ISSUE #819: [Bug] [v5-Cochabamba] Non-intuitive destination selection on map (manual drag required)
**Bug description**
Manually setting a destination on the map is difficult and non-intuitive. The application does not allow a direct "click" to select a location; instead, it forces the user to drag the map until a central marker appears. There are no clear instructions, which creates confusion for the user.

**Severity**
- Priority: Medium 
- Criticality: Medium

Test Environment

- Device: Desktop PC / Laptop
- OS: Windows 10
- Browser: Chrome, Edge (Reproduced in both)

Steps to Reproduce

1. Open the Trufi Web App.
2. Click on the "Destination" (Destino) input field.
3. Try to select a point on the map by clicking directly on a specific location.
4. Observe that nothing happens upon clicking.
5. Drag the map around until the marker/arrow eventually appears in the center.

Evidence

Image

Actual Behavior
The user is forced to manually drag the map to align a central marker with the desired location. Direct clicking on the map does not trigger destination selection.

Expected Behavior
The selection process should be more direct. Clicking anywhere on the map should immediately place the marker at that point or display a "Confirm this point" button.

SamuelRioTz:
Fixes #819.

## Problem

In *"Elegir en el mapa"* the marker is fixed at the centre of the screen and the map only responded to **dragging**. Tapping a place did nothing at all — verified before the change: the pin stayed put and the coordinate chip never moved. For anyone used to a map that responds to a click, this reads as broken rather than as a different interaction model, which is exactly what the reporter described.

## Fix

`onMapClick` now moves the camera so the tapped location sits under the marker, and updates the tracked point. Dragging is untouched — both interactions work.

## Testing

Headless Chrome against the built web app:

- **Before**: tap at another point → chip still `-17.39884, -66.16269`, pin unmoved.
- **After**: same tap → map recentres, chip reads `-17.39563, -66.15604`, pin over the tapped place.

Package suite 37/37 green, analyze clean. Screenshots in the review comment.

### DIFF
--- a/packages/trufi_core_maps/example/lib/engines/flutter_map_engine.dart
+++ b/packages/trufi_core_maps/example/lib/engines/flutter_map_engine.dart
@@ -134,7 +134,7 @@ class _FlutterMapWidgetState extends State<_FlutterMapWidget>
     super.didUpdateWidget(oldWidget);
 
     if (oldWidget.controller != widget.controller) {
-      oldWidget.controller?.detach();
+      oldWidget.controller?.detach(this);
       widget.controller?.attach(this);
     }
 
@@ -149,7 +149,7 @@ class _FlutterMapWidgetState extends State<_FlutterMapWidget>
 
   @override
   void dispose() {
-    widget.controller?.detach();
+    widget.controller?.detach(this);
     _mapCtl.dispose();
     super.dispose();
   }
--- a/packages/trufi_core_maps/lib/src/domain/controller/map_controller.dart
+++ b/packages/trufi_core_maps/lib/src/domain/controller/map_controller.dart
@@ -32,7 +32,17 @@ class TrufiMapController {
   void attach(TrufiMapDelegate delegate) => _delegate = delegate;
 
   /// Called by [TrufiMap] to unbind its state. Do not call directly.
-  void detach() => _delegate = null;
+  ///
+  /// Guarded on purpose: when the map engine changes, Flutter builds the
+  /// new map (which attaches) *before* disposing the old one (which
+  /// detaches). An unconditional detach therefore left the controller
+  /// bound to nothing for the rest of the screen's life — silently
+  /// killing every controller-driven action after a style switch
+  /// (camera moves, marker picking). Only the delegate currently bound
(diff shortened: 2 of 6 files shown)


SamuelRioTz:

| Before — tap does nothing (pin and chip unchanged) | After — tap places the point |
|---|---|
|  |  |

- The screen already tracked the centre through `onCameraChanged`; the tap handler reuses that state and just moves the camera, so there is no second source of truth for the picked point.
- Zoom is preserved across the tap (tracked from the camera instead of assuming the initial value), so tapping doesn't reset how far in the user had zoomed.
- Confirming after a tap yields the tapped coordinates — the flow that feeds #820's reverse geocoding, which returns the street name.

SamuelRioTz:
Fresh-eyes re-review found a **blocking bug that predates this PR** and that the tap fix would have made user-visible. Applied in `f103afcf`:

**`TrufiMapController.detach()` was unconditional.** Flutter builds the new widget before disposing the old one, so switching map style ran `attach(new) → detach(old)` and left the controller bound to **nothing** for the rest of the screen's life. Consequences:

- On this screen: after a style switch a tap updated the coordinates while the marker stayed put — the user would confirm **a different place than the one shown**.
- Already live on the home screen, independent of this PR: the same dead controller silently kills **POI picking** (`pickMarkersAt` returns empty) and the **my-location button** after a style switch.

Fix: `detach(delegate)` only unbinds the delegate that is actually bound; the tap handler also refuses to claim a point when the map isn't bound, so coordinates can never drift from the marker.

Three widget tests now cover it — tap places the point, zoom is preserved, and **the tap still works after `setEngineByIndex`** (that last one fails on the previous commit, which is exactly the regression we want pinned). Package suite 40/40, analyze clean; home_screen and transport_list suites re-run green.

Credit where due: the reviewer built the harness that exposes it (a fake engine reproducing TrufiMap's attach/detach cycle) — adopted here with focused assertions instead of the exploratory probes.

PR #961, opened 2026-08-09
fix(settings): plain-language routing options with a connectivity badge

### LINKED ISSUE #921: [Enhancement][UX][V5] Improve Routing Engine settings to make the options easier to understand
### 📝 Description
The **Routing** settings screen exposes technical terminology that is difficult for regular users to understand.

Currently, users are asked to choose between different routing engines (e.g., **Trufi Planner** and **OTP 2.8.1 OpenTripPlanner Online**). Based on user feedback, many people do not understand what these options mean or which one they should select.

Most users are interested in the result (offline vs. online, accuracy, internet requirement) rather than the underlying routing technology.

**Suggestion:**
Replace technical descriptions with user-oriented explanations that focus on the benefits of each option.

**Current text:**

**Routing**
> Select your preferred routing engine.

- **Trufi Planner**
  - Works offline with GTFS data packaged inside the app.
- **OTP 2.8.1 OpenTripPlanner Online**

**Suggested text:**

Image

* **Priority:** Low
* **Criticality:** Minor - UX / Usability Improvement

### 📱 Test Environment
* **Device:** Android 
* **App Version:** V5.0.6

### 💡 Expected Improvement
Users should immediately understand the difference between both routing options without needing technical knowledge, allowing them to choose the option that best fits their needs.

SamuelRioTz:
Fixes #921.

## Problem

The routing section asked users to pick between **engines**, by name and version:

> **Routing** — Select your preferred routing engine
> · **Trufi Planner** — Works offline with GTFS data packaged inside the app
> · **OTP 2.8.1** — OpenTripPlanner 2.8 (Online)

None of that answers the question people actually have: *will this work without internet, and which one is more up to date?*

## Fix

Applies the copy the reporter mocked up in the issue:

| | Before | After |
|---|---|---|
| Section | Routing — *Select your preferred routing engine* | **Route search** — *Choose the service used to calculate your routes* |
| Trufi Planner | *Works offline with GTFS data packaged inside the app* | *Finds routes using the data saved on your device* + **Works without a connection** |
| OTP | *OpenTripPlanner 2.8 (Online)* | *Finds routes through an online service. May have more up-to-date information* + **Needs an internet connection** |

The connectivity requirement gets its own line with an icon, because that is the real decision criterion; the screen already knew `requiresInternet`, it just wasn't showing it. en/es/de.

The OTP description no longer interpolates the engine version, so the `{version}` placeholder is gone from the ARB and from the three call sites.

## Not included (needs a decision)

The mockup also renames the option itself to **OpenTripPlanner (OTP)**. Both the example and trufi-app pass `displayName: 'OTP 2.8.1'` explicitly, so that rename is an app-side change rather than a core default — happy to do it in the apps if we want the version out of the UI entirely.

## Testing

Analyze clean on both touched packages; settings/routing suites unchanged. Screenshot of the result in the review comment.

### DIFF
--- a/packages/screens/trufi_core_settings/lib/l10n/settings_de.arb
+++ b/packages/screens/trufi_core_settings/lib/l10n/settings_de.arb
@@ -41,6 +41,6 @@
   "limitationRequiresInternet": "Erfordert Internet",
   "limitationSlower": "Langsamere Antwort",
   "limitationNoWalkingRoute": "Keine Fußweg-Route auf Karte",
-  "settingsWorksOffline": "Funktioniert ohne Verbindung",
-  "settingsNeedsInternet": "Benötigt eine Internetverbindung"
+  "settingsWorksOffline": "Ohne Verbindung",
+  "settingsNeedsInternet": "Braucht Internet"
 }
--- a/packages/screens/trufi_core_settings/lib/l10n/settings_de.arb
+++ b/packages/screens/trufi_core_settings/lib/l10n/settings_de.arb
@@ -1,6 +1,5 @@
 {
   "@@locale": "de",
-
   "onboardingTitle": "Willkommen!",
   "onboardingSubtitle": "Lass uns deine Einstellungen festlegen",
   "onboardingLanguageTitle": "Wähle deine Sprache",
@@ -11,7 +10,6 @@
   "onboardingMapTitle": "Wähle deinen Kartenstil",
   "onboardingRoutingTitle": "Wähle deine Routenplanung",
   "onboardingComplete": "Los geht's",
-
   "privacyConsentTitle": "Hilf uns, die App zu verbessern",
   "privacyConsentSubtitle": "Hilf uns, die App zu verbessern, indem du anonyme Nutzungsdaten teilst",
   "privacyConsentInfoTitle": "Was wir sammeln",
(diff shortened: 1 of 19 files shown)


SamuelRioTz:



- The badge reuses `requiresInternet`, which the screen already had — no new state, and it stays correct per platform: the same Trufi Planner shows *Works without a connection* on mobile (local GTFS) and *Needs an internet connection* on web (remote), because that is genuinely what happens.
- The engine version left the description, so the ARB placeholder went with it; the three OTP providers now use a plain getter.
- Screenshot is the web build, hence all three options showing the online badge; on a phone the first one shows the offline badge.

SamuelRioTz:
Fresh-eyes re-review applied.

**#961 (this PR)** — two findings, both fixed:

- **The body contradicted the diff.** It said the `OpenTripPlanner (OTP)` rename was left out pending a product decision, while the code renamed the core default. Reverted: the names stay as they are. The reviewer also pointed out why that rename needs care — an app registering two OTP providers without `displayName` would then show **two identical rows**, same name and same description, which is worse than a version number.
- **The German badge did not fit.** At 320 dp the badge slot is ~146 dp and *"Benötigt eine Internetverbindung"* needs ~190 dp, so it ellipsised away the word it exists to say. Now `Braucht Internet` / `Ohne Verbindung`.

Confirmed clean by the review: the `{version}` placeholder was dropped from all three locales at once (gen-l10n would not break), the generated files are byte-identical to real `flutter gen-l10n` output, all three call sites were updated, and no city app calls that string.

Known and deliberately out of scope: the same choice is still worded differently in the onboarding sheet and in the home-screen routing sheet. Worth a follow-up so all three read the same, but it is copy in other screens rather than this fix.

Issue #745, opened 2024-05-03
Search with an in street intersection


AnivlisLuz:
Currently Google maps has this option, users asks for that.

AnivlisLuz:
Based on the findings from issue #740, there might be a potential solution using Photon. If not, an implementation will be required and will be included as a task in the Peru project.

SamuelRioTz:
The core side of this is now complete and merged:

- #962: offline street/junction data service (`search.json` from `osm-search-data-export`), junction queries ("ayacucho y heroinas"), and the composite service that runs it alongside the online geocoder — they answer different questions, so cities get both.
- #964: the street → corners flow — tapping a street with known corners opens a dedicated sub-screen ("← <street>" title, a filter that searches only within the corners), and picking a corner uses it as the location.

Remaining to close this issue: ship the generated `search.json` in trufi-app (generate → bundle as asset → wire the composite service), which lands with the next core release bump.

SamuelRioTz:
The core side of this is done and merged: street-corner search shipped across #962–#965 (search index tooling, drill-down UI street → corners list with its own filter, ranking). What remains is app-side only — bundling the corners index and deploying — now tracked in trufi-association/trufi-app#42. Closing here: trufi-core is a library, its deliverable ends at the release. Reopen if the feature misbehaves once it reaches the app.

PR #962, opened 2026-08-09
feat(search): offline street and junction search, alongside the online one

SamuelRioTz:
Towards #745 (the street-junction part; the street→corner drill-down UI follows).

## Problem

In Cochabamba — and in most cities Trufi serves — people don't give addresses, they give **intersections**: *"Ayacucho y Heroínas"*. No geocoder answers that. Photon can find a street, or a place, but not the point where two streets cross.

The old Trufi Core had exactly this flow, fed by a `search.json` with streets **and their junctions**. The v5 migration left it behind: today search is Photon-only and the data isn't generated or shipped anywhere.

## What this adds

**The data** — a `search-*` tool in the same shape as the existing `gtfs-*`/`pbf-*` tools, using the association's own [`osm-search-data-export`](https://github.com/trufi-association/osm-search-data-export) as a library:

- `apps/example/tools/search-bolivia-cochabamba/` — Overpass by default so it runs standalone, PBF for repeat runs
- (a sibling tool lands in trufi-app: Cochabamba yields **4025 streets and 23098 junctions**, 2.6 MB)

**The services** — two, designed to work *together* rather than one replacing the other:

- `OfflineSearchDataService` reads `search.json` and understands `A y B`, `A esq B`, `A con B`, `A & B`, case- and accent-insensitive ("heroinas" finds "Heroínas"). It also exposes `junctionsOf(street)` for the drill-down flow.
- `CompositeSearchLocationService` queries several services concurrently and merges the results, deduplicating by name+coordinates. **A service that fails or times out is skipped**, so losing the network degrades to offline street/junction results instead of an error screen.

The example wires both and ships a small central-Cochabamba sample, so the behaviour is visible without running anything.

## Testing

16 unit tests (junction parsing with every separator, accent folding, street/POI matching, no-crossing case, merge order, dedupe, one-source-fails, timeout, reverse fallback, dispose). Analyze clean; the package's preexisting network-dependent failures are unchanged.

End-to-end on the built web example: typing `ayacucho y heroinas` returns the **junction first**, then Photon's results — screenshot in the review comment.

## Still to come for #745

The street→corner UI (pick a street, then choose among its junctions), and shipping the data in trufi-app (generate → copy to assets → version bump).


SamuelRioTz:

Typing `ayacucho y heroinas` on the built example — the junction comes first, Photon's street results follow:



- **Why a composite instead of swapping services**: they answer different questions. Streets and corners are static, local and offline; places, businesses and addresses are better online and stay fresher. Cities that have both should offer both, and the failure mode matters — with the composite, no network means "fewer results", not "search is broken".
- **The junction data is the hard part and it already existed**: `osm-search-data-export` has produced it since the v3 era (`streetJunctions: [streetRef, [lon, lat]]`). This PR doesn't reinvent it, it re-plugs it.
- `reverse()` deliberately returns null in the offline service: the dataset has no addresses, and guessing the nearest street would be worse than deferring to the online one (the composite falls through in order).
- Asset size: the committed example sample is 529 KB for central Cochabamba; the full city is 2.6 MB, which is the number a city app would ship.

SamuelRioTz:
Fresh-eyes re-review — it measured things I had only assumed, and the PR changed shape as a result:

**Offline POIs are gone.** The reviewer benchmarked the full-size Cochabamba file and found the POI half was the weak one: no ranking (`mercado` surfaced *Supermercado Petocho* and *MicroMercado Online* before the actual markets, purely by file order), no dedupe (`plaza` returned **Plaza Colón three times**, all rendered identically because the rows have no address), and it pushed up to 20 offline rows ahead of the first online result. It also cost ~28 ms per keystroke because POI names were normalized inside the scan. Places are the online service's job — this service now owns only streets and junctions, which is what its own doc comment claimed all along.

**Debounce.** The wired path (`LocationSearchScreen.onChanged`) called `search()` on every keystroke with no timer — `ayacucho y heroinas` meant **19 searches and 19 HTTP requests**. (The 300 ms debounce in `SearchLocationsCubit` is on a different path that this screen never takes.) Now 300 ms after typing stops.

**Failures are loud again.** A missing asset escaped as a raw `FlutterError`; `SearchLocationsCubit` only catches `SearchLocationException`, so a city app that forgot the asset in pubspec would show a spinner that never stops. It's wrapped now, and a failed load is no longer cached forever. One malformed junction row used to throw away the entire city's data — bad rows are skipped instead.

Also: street `alternativeNames` are searchable now ("Cbba" finds Cochabamba), the tool creates `out/` and checks the PBF before opening it, and the example's `displayName` removal moved out of this PR — it belongs to #961.

19 tests, now including the missing-asset and malformed-row paths. Behaviour unchanged where it matters:



Correction to my own PR body: the package's preexisting test failures are **binding-initialization** failures in the photon/nominatim suites, not network-dependent ones as I wrote.

PR #964, opened 2026-08-10
feat(search_locations): street → corners drill-down on the search screen (#745)

SamuelRioTz:
Second half of #745 (the first half — offline street/junction data + composite search — landed in #962).

## What

In cities without street numbers people give directions as intersections, so picking a street is rarely the end of a search. The old core let you tap a street and then choose the corner; this restores that flow on top of the offline street data:

- **`SearchLocationDrillDown`** — optional capability mixin for services whose results can be expanded into finer-grained locations. `canDrillDown` is synchronous on purpose: the screen calls it per visible result while building the list.
- **`OfflineSearchDataService`** implements it for street results with at least one well-formed junction. The sync check never races the data load: street results can only exist after the data loaded.
- **`CompositeSearchLocationService`** forwards to whichever child can answer — host apps get the flow with **zero extra wiring**.
- **`LocationSearchScreen`** — street results whose service knows their corners get a trailing fork button. Tapping it swaps the results for the corners list (header with back); picking a corner pops it as the selection; editing the query exits the corners view. The tile body still picks the whole street — both are valid answers.
- **l10n**: `cornersOf` / `seeCorners` in en/es/de.

## Evidence (Android emulator, bundled Cochabamba sample)

`ayacucho` → the offline street shows the fork button (Photon results don't) → *Esquinas de Avenida Ayacucho* (Aroma / Heroínas / Rafael…) → the picked corner lands as the route origin.

## Tests

- `composite_search_location_service_test.dart`: forwarding contract (delegates to the capable child regardless of order; non-drillable results report false without calling children; plain services skipped) — 10/10.
- `offline_search_data_service_test.dart`: street-with-corners drillable, street-without not, junctions not, pre-load answers false — 16/16.
- `location_search_screen_drilldown_test.dart` (new): button only on drillable results; corners list + back; picking a corner pops it; editing the query exits — 4/4.
- `flutter analyze` clean. Pre-existing HTTP-mock test failures (Nominatim/Photon suites) are unchanged vs main. **The CI check runs no tests** — the counts above are local runs.

### DIFF
--- a/packages/trufi_core_search_locations/lib/l10n/search_locations_de.arb
+++ b/packages/trufi_core_search_locations/lib/l10n/search_locations_de.arb
@@ -9,6 +9,5 @@
   "yourPlaces": "DEINE ORTE",
   "searchResults": "SUCHERGEBNISSE",
   "noResultsFound": "Keine Ergebnisse gefunden",
-  "cornersOf": "Kreuzungen: {street}",
-  "seeCorners": "Kreuzungen anzeigen"
+  "searchCornersHint": "Kreuzungen suchen..."
 }
--- a/packages/trufi_core_search_locations/lib/l10n/search_locations_de.arb
+++ b/packages/trufi_core_search_locations/lib/l10n/search_locations_de.arb
@@ -8,5 +8,7 @@
   "chooseOnMap": "Auf der Karte wählen",
   "yourPlaces": "DEINE ORTE",
   "searchResults": "SUCHERGEBNISSE",
-  "noResultsFound": "Keine Ergebnisse gefunden"
+  "noResultsFound": "Keine Ergebnisse gefunden",
+  "cornersOf": "Kreuzungen: {street}",
+  "seeCorners": "Kreuzungen anzeigen"
 }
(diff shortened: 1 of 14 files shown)


SamuelRioTz:
Review feedback applied: rows in the corners list no longer repeat the drilled street's name — they start at "&" ("& Avenida Heroínas de la Coronilla" now reads in full instead of truncating). Display-only: the picked location keeps the complete corner name for the origin/destination field. Widget tests updated accordingly; re-verified on the emulator.

SamuelRioTz:
Second round of review feedback applied — the corners view is now a dedicated sub-screen:

- **Tapping the street row itself opens its corners** (not just the trailing icon): a street with known corners is not a point, so picking it directly answered nothing. Plain results still pop as the selection.
- **Title bar: ← + the street name** — and since the rows below start at "&", the street's name lives only in the title.
- **A filter field right under the title searches only within the corners**, accent-insensitively ("jose" finds "José").
- **Nothing else on that screen** — no quick actions, no saved places.
- The system back leaves the corners first (PopScope), then the screen.

Widget tests reshaped to the new flow (6/6); re-verified on the emulator (screenshots in the ops repo, `prs/pr-964/`).

Issue #847, opened 2026-03-01
[Bug] [v5-Cochabamba] Very long route with unnecessary transfers


Esther2419:
Description of the error

The routing algorithm suggests routes with extremely inefficient transfers instead of using existing direct routes. Even though the route of lines like the 233 Amancayas passes basically right by the destination, the system forces the user to take an initial line and then backtrack on a second one, increasing travel time from 15 minutes to almost 2 hours.

Severity
Medium: Causes inconvenience or annoyance, with temporary solutions available.

Priority: Medium: If it is an error that only occurs in very specific street combinations.

Steps to reproduce

- Open the application on the web https://planner.trufi.app.

- Select a starting point: Pacata School "School-Carvert"

- Select a destination point: "Jardín Martín Cárdenas"

- Observe the route suggestions in the bottom panel.

- See the error on the map where the route drew an unnecessary round trip loop.

Current Result
The application suggests a route combination (e.g., 233 + 46) or the other option (e.g., 233 + 8) that takes 1 hour 5 minutes or up to 1 hour 30 minutes for a journey that appears to be direct.

Expected Result
The algorithm should recognize the proximity of the direct route to the destination and allow the detour, offering the shortest route as the primary option.

Attachments

<img width="1919" height="904" alt="Image" src="https://github.com/user-attachments/assets/ae3dc562-0ed5-4a03-9ac1-7e8553ada0e5" />

![Image](https://github.com/user-attachments/assets/bf004c91-35dc-4e2c-ac42-bb1cb03c925f)

Environment

Device: Mobile/Laptop
Brand: Xiaomi 15T Pro
Operating System: Android/Windows 10
Browser: Chrome Version 145.0.7632.117 (Official Build) (64-bit)
Instance: v5-Cochabamba
Screen: ~6.67"/14 inches

IsmaDavidAA:
hi @Esther2419 good issue, however I need more details, could you share your search please? the link of web browser or link to share route form app, I also need to know which OTP version have this issue, 1.5 or 2.8.0, please share more details

SamuelRioTz:
[...]

PR #965, opened 2026-08-10
fix(routing): order OTP 2.8 itineraries by the router's generalized cost

### LINKED ISSUE #849: [Bug] Routing Priority Error: System promotes slow multi-transfer routes over faster direct options
**Bug description**

A critical failure in the route ranking algorithm has been detected. The application is prioritizing complex routes that require transfers (multiple buses) and longer travel times, while direct routes (single bus) with significantly better arrival times are buried at the bottom of the list.
- The Issue: As shown in the attached evidence, the app first suggests routes taking up to 34 minutes with transfers (Lines LL + 27/Z14/39).
- The Conflict: The most efficient route (Line 212), which takes only 14 minutes and is direct, appears in the 4th position or lower.
- User Impact: A typical user will likely follow the first option presented, causing them to spend twice as much time and pay double the fare (two tickets) unnecessarily.

**Severity**
- Severity: High (The routing engine is failing its core purpose: providing the most efficient path).
- Priority: High.

**Steps to Reproduce**
1. Set a route from "Parque del Maestro" towards the "Hospital Cochabamba" area (Intersection Av. Ayacucho & Aroma).
2. Review the suggested results in the sidebar/panel.
3. Note that the top 3 results require a transfer starting with Line LL.
4. Scroll down to find that Line 212 or Line 233 reaches the destination directly in less than half the time (14 min).

**Actual Behavior**
The system seems to prioritize transfer-heavy routes or fails to calculate the "Total Travel Time" correctly as the primary ranking factor, hiding the most logical option for the citizen.

**Expected Behavior**
Direct routes with shorter total travel times MUST always appear in the first position. Transfers should only be prioritized if they significantly reduce the total duration of the trip.
   
**Attachments**

Image

Image

Image

**Environment**
 - Instance: Cochabamba v5.
 - Device: Desktop PC / Laptop
 - OS: Windows 10
 - Browser: Chrome / Edge

SamuelRioTz:
Fixes #849 · addresses the ranking half of #847.

## The measurement that pinned it down

Against an MD5-identical local copy of the Cochabamba production graph, **Sacaba centro → UMSS** returns:

| OTP's order (what users see today) | duration | generalizedCost |
|---|---|---|
| **233 → 233** (same line, 'Nro. rojo' → 'Bandera verde', a real get-off-and-reboard transfer — not interlining, verified via `interlineWithPreviousLeg`) | 34 min | **3552** |
| 241 direct | 33 min | 2587 |
| 241 direct | 33 min | 2835 |

The router had already priced the transfer as clearly worse — its generalized cost includes the configured `transferPenalty` (30 min) and `boardCost` — but the plan API returns **arrival-time order**, and the app showed it verbatim. That's the entire mechanism behind *"promotes slow multi-transfer routes over faster direct options"*: the config is fine (we also measured that **lowering** the penalties is what reproduces the complaint), the presentation order wasn't.

## What changed

- `Itinerary` now carries `generalizedCost` (query + parser + model, JSON round-trip).
- `Otp28RoutingProvider` re-orders the filtered list by it: **stable** (index tiebreak — `List.sort` is only de-facto stable below ~32 elements), unpriced itineraries (older servers) sink below priced ones keeping their relative order.

## After the fix (same lab)

`241 (2587) → 241 (2835) → 233+233 (3552)` — the direct rides lead.

## Tests

New `otp_28_cost_sort_test.dart` (6/6): the real-world fixture above, stability at 40 equal-cost items, unpriced sinking, null/short passthrough, JSON round-trip. Package suite strictly better than main (79/40 vs 73/41 — remaining failures are the pre-existing network-dependent suites). `flutter analyze`: no new issues. **CI runs no tests** — counts are local.

### DIFF
--- a/packages/trufi_core_routing/lib/src/models/itinerary.dart
+++ b/packages/trufi_core_routing/lib/src/models/itinerary.dart
@@ -17,6 +17,7 @@ class Itinerary extends Equatable {
     this.transfers,
     this.arrivedAtDestinationWithRentedBicycle = false,
     this.emissionsPerPerson,
+    this.generalizedCost,
   }) : legs = _assignDefaultColors(legs),
        distance = _calculateDistance(legs);
 
@@ -29,6 +30,13 @@ class Itinerary extends Equatable {
   final int? transfers;
   final bool arrivedAtDestinationWithRentedBicycle;
   final double? emissionsPerPerson;
+
+  /// OTP's generalized cost for this itinerary (seconds-equivalent):
+  /// duration plus every configured penalty — transfers, boarding, walk
+  /// reluctance. Lower is better *by the router's own judgment*, which
+  /// is exactly what its arrival-time-ordered list does not convey.
+  final int? generalizedCost;
+
   final int distance;
 
   static int _calculateDistance(List<Leg> legs) {
@@ -63,6 +71,7 @@ class Itinerary extends Equatable {
       arrivedAtDestinationWithRentedBicycle:
           json['arrivedAtDestinationWithRentedBicycle'] as bool? ?? false,
       emissionsPerPerson: emissionsData?.getDouble('emissionsPerPersonCo2'),
+      generalizedCost: json.getInt('generalizedCost'),
     );
   }
 
@@ -103,6 +112,7 @@ class Itinerary extends Equatable {
       'arrivedAtDestinationWithRentedBicycle':
           arrivedAtDestinationWithRentedBicycle,
       'emissionsPerPerson': {'emissionsPerPersonCo2': emissionsPerPerson},
+      'generalizedCost': generalizedCost,
     };
   }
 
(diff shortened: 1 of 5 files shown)


SamuelRioTz:
Emulator evidence added (example app pointed at an MD5-identical local copy of the production graph, OTP 2.8 engine, trip "current location (Sacaba centro) → Facultad Tecnología UMSS"):

**Before (main)** — the same-line transfer leads:
1. 32 min · **233 → 233** (transfer)
2. 32 min · 241 direct
3. 32 min · 241 direct

**After (this PR)** — the router's preferred rides lead:
1. 32 min · **241 direct**
2. 32 min · 241 direct
3. 32 min · 233 → 233

Same three itineraries, only the order changes. Screenshots live in the ops repo under `projects/trufi-core/prs/pr-965/`.

Issue #945, opened 2026-08-03
Tracking: remaining fixed-language assumptions (N-language audit)


SamuelRioTz:
A full audit for code that assumes a fixed language set (the class of problem behind #871) found 27 issues. The blocking and high-severity ones are fixed in #944 — this issue tracks the rest so they don't get lost. Each item is independent and PR-sized.

**Context:** Trufi ships in whatever languages a city translates it into (`gen_extra_l10n`, #919). Anything that hardcodes a language, a locale list, or an English/Spanish string breaks that promise. #944 added the infrastructure (fallback delegates, `IRoutingProvider.localized*`, notifications l10n) — these are the remaining consumers.

## Formatting (MEDIA)

- [ ] **Time format is fixed 24h** — `DateFormat('HH:mm')` in ~6 sites (`itinerary_list.dart:633`, `itinerary_card.dart:123,137`, `itinerary_detail_screen.dart` ×4, `leg_ui.dart`, `itinerary_ui.dart`, `share_route_service.dart:68`). Wrong for `en_US` and any language with its own convention. Use `DateFormat.Hm(locale)` / `jm` honoring `MediaQuery.alwaysUse24HourFormat`.
- [ ] **Distance strings hardcode English units and decimal separator** — `'${km.toStringAsFixed(1)} km'` in `transport_detail_screen.dart:779-784`, `otp_24_preferences.dart:481`, `otp_15_preferences.dart:488`. Add `distanceKilometers`/`distanceMeters` keys (transport_list + routing ARBs) and use `NumberFormat`.
- [ ] **Navigation card distance/duration** — `navigation_instruction_card.dart:346-358` (`'2.4 km'`, `'5 min'`, `'1h 20m'`): add `navDistance*`/`navDuration*` keys modeled on `home_screen_en.arb:35-75`.
- [ ] **GTFS frequency display** — `gtfs_frequency.dart:22-32` (`'every 10 min'`): expose the raw `headway` and format in the widget layer; deprecate `displayFrequency` for UI use.

## Wiring (MEDIA)

- [ ] **OTP/geocoder never receive the app language** — `route_planner_cubit.dart:236` doesn't pass `locale:` to `fetchPlan` although the whole chain supports it; `PhotonSearchService`/`NominatimSearchService` accept `language` and no caller passes it (`apps/example/lib/main.dart:260`). Server-provided texts (alerts, geocoder results) come back in the server's default language.
[...]

PR #969, opened 2026-08-10
fix(l10n): forward the app language to routing and geocoding (#945)

SamuelRioTz:
The two **Wiring** items from the #945 tracker.

## The gap

Servers could answer in the rider's language — OTP's `locale` variable and Photon/Nominatim's `lang`/`accept-language` were all plumbed — but **nothing ever told them which language**. Server-produced texts (alerts, geocoder result names/addresses) always came back in the server's default.

## What changed

- **Plan requests**: `RoutePlannerCubit.updateLocale()` carries the app's language tag, kept fresh by the home screen's `didChangeDependencies` (locale is an inherited dependency, so switching languages re-fires it — one wiring point instead of touching all 7 `fetchPlan` call sites), and `fetchPlan` finally passes `locale:` down the already-supporting chain.
- **Geocoders**: new `LanguageAwareSearch` capability, same pattern as `SearchLocationDrillDown` from #964: Photon and Nominatim implement it, `CompositeSearchLocationService` forwards to every capable child, and `LocationSearchScreen` keeps it in sync with the app locale. Language-unaware services are untouched.
- **`TrufiScreen.supportedLocales` removed** — dead API: every screen overrode it, nothing consumed it, and under the N-language model the city's `TrufiLocaleConfig` is the authority on languages (aggregating per-screen lists would contradict #919/#944). Per-screen example apps now use their own localizations' `supportedLocales`.

## Tests

- Composite forwards `searchLanguage` to capable children, skips plain services (unit).
- `LocationSearchScreen` syncs the service language from the app locale (widget).
- Full workspace: **all 11 test-bearing packages green** (`melos ci:test` equivalent, which this repo's CI now actually runs — #966).

### DIFF
--- a/packages/screens/trufi_core_about/lib/about_trufi_screen.dart
+++ b/packages/screens/trufi_core_about/lib/about_trufi_screen.dart
@@ -59,9 +59,6 @@ class AboutTrufiScreen extends TrufiScreen {
   ];
 
   @override
-  List<Locale> get supportedLocales => AboutLocalizations.supportedLocales;
-
-  @override
   ScreenMenuItem? get menuItem => const ScreenMenuItem(
     icon: Icons.info_outline,
     order: 900,
--- a/packages/screens/trufi_core_fares/example/lib/main.dart
+++ b/packages/screens/trufi_core_fares/example/lib/main.dart
@@ -97,7 +97,7 @@ class MyApp extends StatelessWidget {
         GlobalWidgetsLocalizations.delegate,
         GlobalCupertinoLocalizations.delegate,
       ],
-      supportedLocales: screen.supportedLocales,
+      supportedLocales: FaresLocalizations.supportedLocales,
     );
   }
 }
--- a/packages/screens/trufi_core_fares/lib/src/fares_trufi_screen.dart
+++ b/packages/screens/trufi_core_fares/lib/src/fares_trufi_screen.dart
@@ -35,9 +35,6 @@ class FaresTrufiScreen extends TrufiScreen {
   ];
 
   @override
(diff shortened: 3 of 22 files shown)


SamuelRioTz:
Fresh-review pass on this branch caught a **blocker before merge**: forwarding the app language unconditionally sent `lang=es` to Photon, which public deployments reject with **HTTP 400** ("language es is not supported, supported languages are: default, en, de, fr" — verified live against photon.trufi.app and photon.komoot.io). Every search in Spanish — the flagship app's default — died with an exception; the unit tests use fakes, so they couldn't see it.

Fixed in 2b1b378e: `PhotonSearchService` now carries the set of languages its deployment accepts (default `{en, de, fr}`, widenable per city); unsupported app languages are simply not forwarded, which is the pre-#969 behavior. Also from the review: the home screen forwards `languageCode` (OTP's `locale` is a two-letter ISO 639-1 code, never `es-BO`), plus a doc-comment dedup in Nominatim.

Verified OK by the same pass: provider order (the cubit is always above HomeScreen), `updateLocale` runs before the deep-link path so shared links plan with the right language, `$locale` was already declared in the OTP 2.4/2.8 queries (this PR activates it), and no downstream app overrides the removed `TrufiScreen.supportedLocales`.

SamuelRioTz:
Emulator evidence added for the blocker fix: with the pre-fix commit, searching "plaza" in Spanish returns **only offline street rows to the very end of the list — not a single Photon place** (the composite swallows the 400); with the fix, the same scroll shows Photon's places (Plaza Libertad, Sucre, Avaroa… with address subtitles). Screenshots in the ops repo (`prs/pr-969/screenshots/`).

PR #970, opened 2026-08-10
fix: fresh-review follow-ups — NumberFormat locale fallback + OTP 1.5 endpoint path check

SamuelRioTz:
Two real findings from the adversarial re-review of the merged #966/#968 batch (process: the fresh-eyes audit that the review flow mandates — run late this time, which is on us, but run).

## 1. `NumberFormat` crashes on N-language locales (latent, from #968)

`NumberFormat('#,##0.0', 'qu')` throws `ArgumentError: Invalid locale "qu"` — verified against intl 0.20.2's `numberFormatSymbols` table (118 locales, no `qu`/`ay`/`gn`, no `'fallback'` entry). The #944 ARB/Material fallback machinery doesn't help: intl is a separate layer. The day a city ships Quechua or Aymara — the exact languages the N-language model exists for — the route-detail stats, both OTP walk-distance sliders and the navigation card red-screen.

**Fix**: `safeNumberFormat()` in `trufi_core_utils` — try the locale, fall back to the pattern's default (same pattern `_localeDateFormat` already uses). `trufi_core_routing` gets a private copy of the guard (it doesn't depend on utils). Regression test formats with `'qu'` and `'ay'`.

## 2. `Otp15RoutingProvider._restEndpoint` matches '/plan' inside the *hostname* (pre-existing, surfaced auditing #966)

`'https://plan.example.com'.contains('/plan')` is true (`//` of the scheme + the host), so the endpoint was returned verbatim without appending the REST path — broken routing for any deployment whose hostname starts with "plan". Not introduced by #966 (which only moved the slash trim); the sibling `_buildIndexEndpoint` already used the correct path-based check. Now decided on `Uri.path`. Regression test pins `plan.example.com → /otp/routers/default/plan`.

## Verification

utils 32/32 (incl. new qu/ay test) · routing 90/90 (incl. new host test) · transport_list 25/25 · navigation 7/7 · analyze clean (only pre-existing infos).

### DIFF
--- a/packages/screens/trufi_core_transport_list/lib/src/transport_detail_screen.dart
+++ b/packages/screens/trufi_core_transport_list/lib/src/transport_detail_screen.dart
@@ -7,7 +7,7 @@ import 'package:latlong2/latlong.dart';
 import 'package:trufi_core_base_widgets/trufi_core_base_widgets.dart';
 import 'package:trufi_core_maps/trufi_core_maps.dart';
 import 'package:trufi_core_routing/trufi_core_routing.dart';
-import 'package:intl/intl.dart';
+import 'package:trufi_core_utils/trufi_core_utils.dart';
 
 import '../l10n/transport_list_localizations.dart';
 import 'models/transport_route.dart';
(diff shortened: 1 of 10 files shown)


SamuelRioTz:
Emulator evidence added — the example app already offers Runasimi (qu) in its language picker, so the bug photographs live: **on current main, switching to Runasimi and opening any route detail red-screens with "Invalid argument(s): Invalid locale 'qu'"**; with this PR the same flow renders "27,1 km · 41 Paradas" normally. Screenshots in the ops repo (`prs/pr-970/screenshots/`).

PR #976, opened 2026-08-11
Widen the planner's stop candidate pool: 60 → 150

### LINKED ISSUE #974: Local planner transfer search matches patterns, not routes — fragmented feeds lose valid transfers
Found while diagnosing trufi-sanaa#2 (Sana'a airport → downtown returns no routes even at a 1500 m walk radius).

## Symptom

A valid one-transfer trip exists in the feed — board line **14** near the airport, transfer at a shared stop, continue on line **7** to downtown — and `GtfsRoutingService` returns nothing.

## Cause

`_findOneTransferRoutes` prunes with `relevantDestStops[conn.otherPatternId]`: the connection's *pattern id* must be one of the patterns observed at the destination-side candidate stops. In fragmented feeds (one pattern per trip variant — Sana'a's line 7 has 5+ patterns), the pattern that shares the transfer stop with line 14 and the pattern that serves the destination stops are **different objects of the same route**, so every candidate dies at that check.

Instrumented funnel on the real feed (origin = airport area, dest = downtown, radius 1500 m, 60 candidates/side):

```
connections scanned:            2352
after direction filter:         1260   ← all pointing at route-7 patterns
after relevantDestStops check:     0   ← none of those pattern ids serve a dest stop
```

…while 5 distinct route-7 patterns DO serve the destination stops (they're just not the ones connected at the transfer stop). The direct phase has the same fragility (same-pattern requirement), which also hid a route-14 stop 75 m from the destination.

## Direction

The transfer (and direct) matching should tolerate pattern fragmentation — e.g. match at the **route+direction level** and then resolve a concrete pattern that contains both stops, or merge near-identical patterns at index build time (patterns whose stop sequences are subsets of one another). Secondary: with `maxStopCandidates = 60`, a dense downtown query only sees stops within ~400 m even when the radius allows 1500 m — worth scaling candidates with radius.

Repro data: trufi-sanaa's bundled `sanaa.gtfs.zip`, origin (15.476, 44.220), dest (15.347, 44.205).

SamuelRioTz:
Fixes #974 — with the corrected diagnosis measured in [this comment](https://github.com/trufi-association/trufi-core/issues/974#issuecomment-corrected): per-variant patterns are semantically right; the real bug was the candidate pool truncating dense-downtown searches to a ~400 m ring regardless of the walk radius.

## The change

One default, everywhere it's declared: `maxStopCandidates` 60 → 150 (config, clients, service — already per-app tunable via `TrufiPlannerConfig`).

## Measured (real feeds)

| | before (60) | after (150) |
|---|---|---|
| **Sana'a** airport → downtown (r=1500) | 0 paths | **1 path: 14→7 transfer**, 4 ms |
| **Cochabamba** 60 random pairs (657 patterns) | avg 2.8 ms · max 19 ms · 162 paths · 16 routeless | avg **3.4 ms** · max **8 ms** · **172 paths** · **14 routeless** |

The worst case *improves*: a wider pool finds directs sooner, and directs suppress the transfer enumeration.

## Scenario battery (new, `stop_candidate_pool_test.dart`)

Per review request: several performance-guarding tests across distinct network shapes, with deliberately loose ceilings (they catch order-of-magnitude complexity regressions, not machine variance — actuals are printed to the log):

1. **The Sana'a shape** — noise-dense center hides the connecting line's boarding stop beyond rank 60: defaults must find the A→B transfer, and a companion test keeps the 60-pool failure as executable documentation of #974.
2. **Dense grid** (Cochabamba shape): 1,600 stops / 80 crossing routes, 30 cross-city queries — all connected, avg 2.9 ms, max 15 ms locally.
3. **Fragmented corridor** (route-212 shape): 40 near-identical trip variants — direct found, 3 ms.
4. **Star hub** (worst-case connection density): 60 routes through one stop, both directions — every spoke pair connects, avg 0.9 ms.
5. **Radius interplay**: larger radius never yields fewer paths.

Suites: planner 12/12, routing 90/90 (integration-tagged network tests excluded, as in CI), analyze clean (two pre-existing infos on main untouched).

### DIFF
--- a/packages/trufi_core_planner/lib/src/client/local_planner_client.dart
+++ b/packages/trufi_core_planner/lib/src/client/local_planner_client.dart
@@ -72,7 +72,7 @@ class LocalPlannerClient implements PlannerRoutingClient {
     required LatLng destination,
     double maxWalkDistance = 800,
     int maxResults = 5,
-    int maxStopCandidates = 60,
+    int maxStopCandidates = 150,
   }) async {
     if (_routingService == null) return [];
     return _routingService!.findRoutes(
--- a/packages/trufi_core_planner/lib/src/client/planner_routing_client.dart
+++ b/packages/trufi_core_planner/lib/src/client/planner_routing_client.dart
@@ -17,7 +17,7 @@ abstract class PlannerRoutingClient {
     required LatLng destination,
     double maxWalkDistance = 800,
     int maxResults = 5,
-    int maxStopCandidates = 60,
+    int maxStopCandidates = 150,
   });
 
   /// Get all transit routes.
--- a/packages/trufi_core_planner/lib/src/client/remote_planner_client.dart
+++ b/packages/trufi_core_planner/lib/src/client/remote_planner_client.dart
@@ -62,7 +62,7 @@ class RemotePlannerClient implements PlannerRoutingClient {
     required LatLng destination,
     double maxWalkDistance = 800,
     int maxResults = 5,
-    int maxStopCandidates = 60,
+    int maxStopCandidates = 150,
   }) async {
     final response = await _httpClient.post(
       Uri.parse('$_baseUrl/plan'),
--- a/packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart
+++ b/packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart
@@ -178,7 +178,7 @@ class GtfsRoutingService {
     int maxTransfers = 1,
     int maxDirects = 5,
     int maxTransferPaths = 5,
-    int maxStopCandidates = 60,
+    int maxStopCandidates = 150,
   }) {
     // Consider every stop within walking distance (bounded by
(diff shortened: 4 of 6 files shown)


SamuelRioTz:
House-flow evidence — the motivating case on the real Sana'a app (same deep-linked origin/destination, only the core underneath changes):

| core `main` (pool 60) | this PR (pool 150) |
|---|---|
|  |  |
| `Found 0 paths` → 'no routes' | `Found 1 paths`: **line 14 → line 7**, 1 h 21 m, 17.6 km, one transfer |

Adversarial fresh review ran with performance as the explicit focus — verdict after its notes were applied: the 'worst case improves' claim was softened to honest numbers (the direct phase is superlinear in the pool; densest pairs can ~double, ~20→~50 ms desktop, single-digit averages), and it uncovered a pre-existing spatial-index bug that made the old pool semi-arbitrary — filed as #977 (with it fixed, the pool's coverage becomes deterministic). Follow-up notes recorded on #974.

PR #978, opened 2026-08-11
Spatial index: k-NN eviction respects distance order

### LINKED ISSUE #977: GtfsSpatialIndex k-NN returns wrong neighbors — eviction ignores order
Found by the adversarial review of #976, verified against brute force on the real Sana'a feed.

`_KdTree._searchNearest` (`gtfs_spatial_index.dart:106-114`) accumulates results **unsorted** until the list fills: `results.last` is not the worst element, so the first eviction throws out an arbitrary stop and the pruning bound `worstDist` is wrong from then on.

Measured (dense downtown point, Sana'a feed):
- k=60: kd-tree max distance 1,400 m vs true 401 m — **34 of the true 60 nearest are missing (57% wrong)**
- k=100: 0 missing · k=150: 1 missing
- Reach is non-monotonic in k (60 → 1,400 m, 80 → 480 m, 120 → 1,488 m)

Consequences: the candidate pool was never 'the k nearest' but a semi-arbitrary subset — part of why #974's symptoms were so erratic, and part of why the wider pool in #976 helps (a big k masks the bug). With a correct k-NN the pool's coverage becomes deterministic and the #976 margin is real.

**Fix is small**: keep the working set ordered (sorted insertion or track-the-max) so eviction removes the true worst and the pruning bound is valid. Add a property test against brute force on a few hundred random points.

SamuelRioTz:
Fixes #977 (found by the adversarial review of #976).

## The bug

`_KdTree._searchNearest` accumulated results **unsorted** until the list filled; from then on it evicted `results.last` — an arbitrary element — and used its distance as the pruning bound. So `findNearestStops(k)` returned a semi-arbitrary subset of the k nearest, and everything built on the candidate pool (#974/#976) inherited erratic coverage.

Measured on the real Sana'a feed (downtown point, brute-force ground truth):

| | before | after |
|---|---|---|
| k=60 missing true neighbors | **34 of 60** (reach 1,400 m vs true 401 m) | **0** (reach 402 m) |
| k=150 missing | 1 | **0** (reach 762 m ≙ true 759 m) |
| reach monotonic in k | no (60→1,400 m, 80→480 m) | yes |

## The fix

Keep the result list sorted from the moment it fills (one sort), then maintain the invariant with binary-search insertion on every eviction — which also **removes the full `sort()` the old code ran per eviction**. Eviction now discards the true worst and the pruning bound is valid.

## Tests

`spatial_index_knn_test.dart`: property test vs brute force (100 query points × k∈{10,60} over a clustered 2,000-stop field, same haversine as the index), reach monotonicity across k, and result ordering. **All three fail on `main`** (verified by stashing the fix); planner suite 15/15 with it.

With this, #976's pool of 150 becomes deterministic — the pool now truly is the 150 nearest.

### DIFF
--- a/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
@@ -104,12 +104,29 @@ class _KdTree {
     final dist = _haversineDistance(lat, lon, node.stop.lat, node.stop.lon);
 
     if (dist <= maxDistance) {
+      // Invariant: once full, `results` is sorted by distance so
+      // `results.last` is the true worst — both the eviction and the
+      // pruning bound below depend on it. The previous code accumulated
+      // unsorted and evicted `.last` blindly, so the pool could keep an
+      // arbitrary subset instead of the k nearest (#977).
       if (results.length < maxResults) {
         results.add(_NearestResult(stop: node.stop, distance: dist));
+        if (results.length == maxResults) {
+          results.sort((a, b) => a.distance.compareTo(b.distance));
+        }
       } else if (dist < results.last.distance) {
         results.removeLast();
-        results.add(_NearestResult(stop: node.stop, distance: dist));
-        results.sort((a, b) => a.distance.compareTo(b.distance));
+        var lo = 0;
+        var hi = results.length;
+        while (lo < hi) {
+          final mid = (lo + hi) >> 1;
+          if (results[mid].distance <= dist) {
+            lo = mid + 1;
+          } else {
+            hi = mid;
+          }
+        }
+        results.insert(lo, _NearestResult(stop: node.stop, distance: dist));
       }
     }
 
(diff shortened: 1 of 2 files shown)


SamuelRioTz:
Emulator evidence (house rule: every PR gets photographed). Honest framing first: **at shipped defaults this fix's user-visible delta on our two real feeds is nearly nil** — an exhaustive sweep (240 seeded pairs × both feeds × k∈{60,150}) found exactly **one** route-list difference; the fix's real value is determinism (property-tested), robustness on other feeds/configs, and the measured 2.5× speedup on eviction-heavy queries. That one visible case, photographed (example app + full Cochabamba feed, same deep-linked pair, only the kNN underneath changes):

| main (broken kNN) | this PR |
|---|---|
|  |  |
| top option **107 → 19** (918 m walk) | top option **107 → A** (927 m, correct nearest-stop set → correct score order) |

The candidate pool now being *actually* the k nearest is what reorders the tie — small here, structural everywhere.

PR #979, opened 2026-08-11
Content-aware offline extraction + shared tile sets

SamuelRioTz:
Refines #975 after the maintainer's design review of the update story: *most app updates don't touch map assets — the engine should detect whether re-extraction is needed at all*.

## What changes

**1. Content fingerprint, gated by build.** The `.extracted-for` marker now stores `build\nfingerprint`. Routine boots read one marker and do nothing else (no hashing — the design constraint). The first boot after an update fingerprints the bundled assets (reads only, memoized per asset per process); identical fingerprint → keep everything, restamp, done. Only real asset changes wipe and re-extract.

**2. One tile set per asset, not per engine.** Visual styles share their mbtiles; extracting a copy per style quadrupled cache and update cost. Tiles now live in `offline_maps/_shared/<asset>` with a content-hash sidecar; styles point at the shared file; legacy per-engine copies are deleted on first contact.

## Measured (example app, 4 styles over one 25.8 MB mbtiles, emulator; baseline boot 12 s)

| scenario | v5.22 today | this PR |
|---|---|---|
| routine boot | baseline | baseline (unchanged) |
| update, maps untouched *(the common case)* | full re-extraction: ~107 MB of writes | **+4 s of reads, zero writes** |
| update with new maps | ~107 MB | **+7 s, ~26 MB** (one copy) — verified: shared file's md5 == new bundle |
| permanent cache | **107 MB** | **33 MB** |

## Tests (6, in `offline_engine_refresh_test.dart`)

First-boot stamp+layout · same-build fast path (documents that assets aren't rehashed) · **update-without-changes keeps extraction and restamps** (sentinel file proves no rewrite) · update-with-changes re-extracts · legacy single-line-marker + per-engine-tiles migration · two engines share one tile copy and both styles reference it. Full maps suite 50/50; analyze clean.

### DIFF
--- a/packages/trufi_core_maps/lib/src/configuration/map_engine/offline_maplibre_engine.dart
+++ b/packages/trufi_core_maps/lib/src/configuration/map_engine/offline_maplibre_engine.dart
@@ -112,13 +112,21 @@ class OfflineMapLibreEngine implements ITrufiMapEngine {
   @visibleForTesting
   static void resetBundleHashCacheForTesting() => _bundleHashMemo.clear();
 
+  /// In-band token for "this asset is not in the bundle". Safe by
+  /// construction: real hashes are lowercase hex, and this token contains
+  /// non-hex letters, so it can never collide with one. It is data, not
+  /// control flow — the only consumer concatenates it into the
+  /// fingerprint, where absence must be representable so that adding or
+  /// removing an asset between builds changes the fingerprint.
+  static const _missingAssetToken = 'missing';
+
   static Future<String> _bundleAssetHash(String assetPath) {
     return _bundleHashMemo.putIfAbsent(assetPath, () async {
       try {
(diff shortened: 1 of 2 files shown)


SamuelRioTz:
On-device rendering evidence (the unit tests verify files; these verify **MapLibre actually renders from the deduplicated `_shared` path** on a real device — example app, branch build):

| Offline Liberty | Offline Dark Matter |
|---|---|
|  |  |

Two different visual styles, **one shared 25.8 MB tile file** — full label/road rendering in both. Measured on the same device: update without map changes +4 s (reads only), with new maps +7 s, permanent cache 107 → 33 MB. Fresh-review findings (injective shared key, utf8 fingerprint) applied and re-verified — suite 50/50, CI green on both pushes.

PR #981, opened 2026-08-12
Home: live-vehicle info panel on tap + configurable initial toggle state

SamuelRioTz:
## What

1. **Tapping a live bus marker now opens an info panel** (route, plate, LIVE badge, speed, last-report age). The layer was display-only: markers ignored taps even though the map's `pickMarkersAt` machinery existed (only POIs used it). Vehicle markers are checked first in `_onMapClick` — picking sorts by distance, so this ordering is what gives realtime info tap priority.
2. **`HomeScreenConfig.liveVehiclesInitiallyEnabled`** (default `false` — no behavior change): lets a deploy ship with the live layer on for first boots. A user's persisted toggle choice still wins. Motivation: the Lima operator pilot, where realtime is the headline feature and a fresh install showing no buses reads as broken.
3. **Real per-vehicle freshness** (added after adversarial review): the OTP GraphQL query now requests `lastUpdate`/`lastUpdated`, so `VehiclePosition.timestamp` is the vehicle's own report time from the GTFS-RT feed instead of the poll time.

## How

- `LiveVehicleInfoPanel` is a public, self-contained widget → widget-testable without a map engine.
- Speed converts from the GTFS-RT wire unit (m/s) to km/h.
- Route ids arrive as `<feedId>:<id>` — the feed prefix is stripped for riders.
- 6 new l10n keys (en/es/de) + regenerated localizations.

## Tests

- 10 new tests: panel rendering (route/plate/badge/speed conversion/freshness strings/future-clamp), graceful degradation, feed-prefix stripping, config default, and 4 pinning the feed-timestamp parsing (ISO, epoch fallback, bad-ISO fallback, absent→poll time).
- Full suite: `melos ci:test` green locally (11 packages); `flutter analyze` clean (pre-existing infos only).

## Evidence (emulator, example app against a live OTP 2.9 feed)

- Tap on bus → panel 'Ruta 1132 / DEMO-1132-R2 / EN VIVO / 25 km/h / Actualizado hace 0 s' (es locale) — the age tracks the vehicle's report time.
- Tap on empty map → no panel, POI behavior unchanged.
- `liveVehiclesInitiallyEnabled: true` → layer already on at first boot.

## Note for reviewers

Unrelated limitation found while harnessing, worth its own issue: two `Otp28RoutingProvider` instances collide because `id` is hardcoded `'otp_2_8'` — the engine manager keeps only one.


SamuelRioTz:
**Adversarial fresh-review findings, addressed in e19466d:**

1. **BLOCKER — freshness was fiction.** `OtpVehiclePositionsProvider` stamped `DateTime.now()` at parse time and the GraphQL query never requested the feed's timestamp, so the panel would report 'updated ≤10 s ago' forever, even for a parked bus with a dead GPS. The query now requests `lastUpdate` (ISO-8601; verified against [OTP's schema docs](https://docs.opentripplanner.org/api/dev-2.x/graphql-gtfs/types/VehiclePosition) and live against an OTP 2.9 server) with the deprecated epoch `lastUpdated` as fallback, parsed by a `@visibleForTesting` helper pinned by 4 new tests.
2. **REAL — tap picking could be crowded out.** `pickMarkersAt` sorts by distance (not z-order) and the previous `globalLimit: 3` let stacked POI/route markers push the bus out of the picked list. The limit is gone; only `vehicle-`-prefixed markers are considered. The PR description's 'vehicles render on the top layer' claim was also wrong (route markers render above them at layerLevel 2000) — priority comes from checking vehicles first in `_onMapClick`, and the comments now say so.
3. Minor: future timestamps (clock drift) clamp to '0 s' with a test; stale comment fixed; test count corrected below.

Re-verified on emulator against the live feed after the fix: the panel's age now tracks the vehicle's own report time.

Remaining observations from the review (panel is a static snapshot while open; no hour unit for very old ages; `Otp28RoutingProvider.id` collision) noted for follow-up issues, not blockers.

PR #982, opened 2026-08-12
Fix: live vehicles vanished during search in hybrid deploys (feed-prefix mismatch)

SamuelRioTz:
## Bug

In a hybrid deploy (bundled-GTFS routing + OTP-fed vehicles — the Lima pilot), starting a route search made **every live bus disappear** and killed the itinerary's LIVE badge. Reported live on the Lima app.

Cause: when results are shown, the home screen filters vehicles by the itinerary's route ids. Legs from the bundled planner carry bare ids (`1132`) while OTP's GraphQL emits feed-prefixed ones (`1:1132`); strict equality matches nothing. Regression from switching the live layer to OTP consumption (previously the endpoint-direct JSON used bare ids, which happened to match).

## Fix

`stripGtfsFeedPrefix()` in `trufi_core_interfaces` as the single normalization point, used by:
- the provider query extensions (`vehiclesForRoute`, `vehiclesForRoutes`, `hasDataForRoute` — the LIVE badge's source),
- the home screen's search-time vehicle filter,
- the info panel's `displayRoute`.

Within one deploy all feeds belong to one city, so bare-id collisions across feeds are not a practical concern (documented in the helper).

## Tests

5 new tests pin matching in both directions (bare↔prefixed, mixed sets, badge path, null routeId). Full workspace suite green (12 packages).

## Evidence

Lima app (local core override, checksum-verified install) against the production feed: with an active search, the 1132 buses render along the corridor **filtered to the itinerary's route**, and the LIVE dot is back on the leg pill. Zero exceptions.

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -1526,13 +1526,20 @@ class _HomeScreenState extends State<HomeScreen>
     if (_realtimeVehiclesProvider != null &&
         _liveVehiclesEnabled.value &&
         _liveVehicles.isNotEmpty) {
-      final toShow = _selectedTransitRouteIds.isEmpty
+      // Feed-agnostic matching: itinerary legs may come from a bundled-GTFS
+      // planner (`1132`) while vehicles come from OTP (`1:1132`) — strict
+      // equality here made every bus vanish the moment a search was active
+      // in hybrid deploys (Lima pilot regression, caught by Sam).
+      final selected = _selectedTransitRouteIds
+          .map(stripGtfsFeedPrefix)
+          .toSet();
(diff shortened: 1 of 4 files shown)


SamuelRioTz:
Adversarial fresh-review verdict: **no blockers**; one real finding, addressed — the LIVE-badge test was vacuous (the mixed fixture gave strict equality a literal twin for every query, so `hasDataForRoute` could regress silently). New assert queries a bare id whose vehicle only exists feed-prefixed; verified to fail under main's strict equality. Review also verified: no production caller depends on prefixed strict equality, Trujillo-style all-OTP deploys keep matching, `displayRoute` is byte-identical on every edge, and 2 of the 5 tests fail on main (non-vacuous). Noted for follow-up: `trufi_planner_data_source.dart:100` uses `split(':').last` — a third normalization semantics that predates this PR.

PR #983, opened 2026-08-12
Group near-duplicate transfer itineraries with interchangeable route options (#737)

### LINKED ISSUE #737: CBBA: Enhance Route Variant Differentiation
Based on user feedback, improve the distinction between route variants by incorporating emojis or flags into their names. This modification aims to enhance user experience by making it easier to recognize and select the desired route quickly.

SamuelRioTz:
Closes #737.

## What

Transfer searches return many combinations that are really **one decision — board the same main bus** — with interchangeable connections (finish with B, C or D). Each combination used to render as its own row. Now they collapse Google Maps style:

- One row per group, keyed on the **main (first) bus**: same route, boarding within walking tolerance. The legs after it are the group's options — different main routes never merge, even from the same stop.
- Joined route chips capped at two: `P → 106 / 120`, and beyond that `106 / 120 +n` — the expansion lists the rest.
- A `+N options` / `+N more` badge expands to the concrete alternatives, each stating its ride and times.
- **The detail view has an option switcher**: chips (`P → 106 · 18:31`) flip the detail between the group's members and select them, so the map follows.
- **Product cap: at most two vehicles per trip** (`filterMaxTransitLegs`, fail-open when nothing passes). Chains like `18 → 120 → 50` are gone; on the real OTP 1.5 dataset this alone turns 15 raw itineraries into 6 rows.

The grouped-list UI (model, render, l10n) existed since `ItineraryGroup` landed but **nothing ever assigned `Plan.groupedItineraries`** — the UI always fell back to the flat list. This PR adds the missing producer and wires it for **all providers** (trufi_planner, OTP 1.5, OTP 2.4, OTP 2.8) in the cubit, post-fetch and post-restore.

## How

- `groupItineraries()` (new, `trufi_core_routing`): same transit-leg count + same main leg — feed-prefix-agnostic route match (`stripGtfsFeedPrefix`, the #982 lesson) with 500 m boarding/alighting tolerance (calibrated against real OTP 1.5 plans over the Cochabamba GTFS). Unidentifiable main routes fail closed.
- Ranking untouched (#965): groups keep the incoming order of their best member; `groups.first.representative` is always `itineraries.first`.
- `ItineraryGroup.slotRoutes`: distinct routes per slot across the group — what the card joins as `106 / 120`.
- `RoutingEngineManager.maxItineraries` stays 5 by default; raising it is per-app opt-in (the example passes 15). OTP2 already filters near-duplicates server-side (`groupSimilarityKeepOne`), so a global bump would buy OTP2 deploys ~3× payload, not better grouping.

## Example app: fully serverless

- `OTP15_ENDPOINT`/`OTP28_ENDPOINT` dart-defines + `tools/local-otp/README.md` runbook → both OTP engines run on localhost over the bundled GTFS; the local planner already worked offline.
- The bundled GTFS carried `route_color: forestgreen` (CSS name, invalid per spec). OTP 2.x tolerates it, **OTP 1.5 dies at graph build** (`NumberFormatException`). Normalized to hex; same rendered colors, no route/stop data changed.

## Tests

Routing 112 (grouper semantics incl. main-bus rule, fail-closed unnamed routes, two-vehicle cap with fail-open, real OTP 1.5 fixture collapsing 5→3 groups) + home screen 27 (cubit fetch/restore wiring, vehicle cap, pinned-time behavior, 320dp zero-overflow, collapse semantics, detail switcher). Full `melos ci:test` + `ci:analyze` green locally — **CI runs no tests**.

## Emulator evidence (fully serverless: local planner + OTP 1.5 on localhost)

Valle Hermoso → Matadero Tiquipaya over OTP 1.5: 6 rows (was 15 raw), first card `P → 106 / 120` with `+1 opciones`; expanding lists `P → 120` with its own times; opening the detail shows the option chips and switching to `P → 120` re-renders header, timeline and map selection. Same-route groups show `+1 más` → "Otras salidas". The local planner's 3 combinations correctly stay ungrouped (different main-bus corners). Zero exceptions in logcat.

## Review history

Three adversarial fresh-review passes (see comments): #1 found 2 MAJOR + 6 MINOR (fixed in 472b34b4), #2 mutation-tested the fixes and caught a layout regression + a pre-existing 12-hour-locale header overflow (fixed in 0585bea1), #3 audited the product-feedback commit (f0920416).

## Follow-ups (non-blocking)

- Pre-existing `RangeError` via shared deep link with negative `route` index — separate issue.
- The option switcher lives in the inline detail; apps overriding `onItineraryDetails` with custom navigation don't get it automatically.
- Request-side transfer caps (OTP `maxTransfers` param) could save server work that the client filter now discards.


SamuelRioTz:
An adversarial fresh-review pass over this PR found 2 MAJOR + 6 MINOR issues; all are addressed in 472b34b4:

**MAJOR — pinned-time deploys (trufi-app) got blank alternative rows.** With `routingTimeOverride` set, clock times are hidden by design, so same-route "other departures" rendered as an icon and a chevron — nothing else. Same-route groups now offer no expansion under pinned time (their members are indistinguishable to the rider), and any row that does show states its ride ("123 → 120") whenever times are hidden. Pinned by widget tests.

**MAJOR — `maxItineraries` 5→15 reverted; the pool is now opt-in per app.** OTP2's server-side filter chain (`groupSimilarityKeepOne`, default 0.85) already discards the near-duplicates client-side grouping would collapse — so the global bump bought OTP2 deploys (e.g. Trujillo, which constructs the manager without an override) roughly 3× payload and more distinct rows, not better grouping. The manager doc now spells out when raising it pays off (OTP 1.5/2.4, local planner — providers without server-side dedup); the example passes 15 explicitly. Reference: OTP RouteRequest docs on itinerary filtering; opentripplanner/OpenTripPlanner#2936; the client-side precedent for this feature is otp-react-redux's `mergeItineraries` (PR #1376), which matches this design.

**MINOR fixes** — selecting a non-representative alternative now forces its group open (collapsing it left the list with no highlighted row while the map drew that alternative); "N routes found" counts grouped rows; nameless-route feeds can't flip a group to "+N options"; walk-only and bike-only no longer merge (zero-slot equivalence requires the same non-transit mode set); no live badge on multi-route chips (it only reflected the representative's route); footer badge and alternative-row times ellipsize instead of overflowing; alternatives with equal start times keep ranked order (Dart's sort is not stable).

**New coverage** for the wiring the original suite missed: deleting the cubit's grouping hook now fails tests (fetch and restore paths — JSON drops the groups), plus widget tests for the pinned-time badge suppression and expanded-row route summaries.

Full `melos ci:test` and `ci:analyze` green locally (reminder: CI runs no tests). A second fresh-review pass over the fix commit is running; results will be posted here.

SamuelRioTz:
Second fresh-review pass (over the fix commit itself) verified 8 of the 10 fixes clean — including mutation-testing the new tests: reverting the `expandable` gate or the grouper changes fails exactly the tests that pin them. It also caught that the overflow fix had introduced a real layout regression; addressed in 0585bea1:

- **Times truncated on ordinary phones**: loose `Flexible` time texts split the free space evenly with the tight `Spacer`, capping the departure time at ~a third of its width ("Depar…") — universally, not just on narrow screens. Times now own the remaining width (`Expanded` + `FittedBox.scaleDown`; scaling beats mid-word ellipsis for clock times).
- **Footer badge lost its right alignment** and ellipsized with room to spare (same mechanism) → `Expanded` + `Align(centerRight)`.
- **Bonus, pre-existing on `main`**: the card header overflowed on 12-hour locales ("8:00 AM → 9:00 AM") at any ordinary phone width — same FittedBox treatment. The grouped list now renders a 320dp phone with zero RenderFlex overflows, pinned by a widget test.
- **Dead chevron**: collapsing a group held open by its selected alternative toggled the expansion Set but the force-open won the next build (and desynced the Set). The collapse tap now moves the selection back to the representative so the group really closes — also pinned by a test.

Full suite green locally (routing 141, home screen 24; `melos ci:test` + `ci:analyze`).

SamuelRioTz:
**Verdict: ✅ Approved** (self-authored PR — posting the review as a comment; merge on maintainer's go)

## Code review

- The consumer side of grouping (`ItineraryGroup`, the expandable grouped render, l10n) has existed on `main` since the model landed, but nothing ever assigned `Plan.groupedItineraries` — verified by grep, the only reader was `itinerary_list.dart` and there were zero writers. This PR adds the missing producer and wires it in the cubit (post-fetch and post-restore), so it covers all four providers.
- Segment-based equivalence with distance tolerances (500 m same route / 300 m across routes) **calibrated against real OTP 1.5 plans** over the bundled Cochabamba GTFS; union-find absorbs the non-transitive metric; route ids compare feed-prefix-agnostically (the #982 lesson). Ranking (#965) is untouched: `groups.first.representative` is always `itineraries.first`.
- Matches the only serious client-side precedent in the ecosystem: otp-react-redux's `mergeItineraries` (opentripplanner/otp-react-redux#1376). OTP2 already filters near-duplicates server-side (`groupSimilarityKeepOne` — RouteRequest docs, opentripplanner/OpenTripPlanner#2936), which is why `maxItineraries` stays 5 by default and raising it is per-app opt-in.
- No overlaps: only open PR in the repo.
- Tests: full `melos ci:test` + `ci:analyze` green locally — routing 141 (grouper unit + a real OTP 1.5 fixture collapsing 5→3 groups), home screen 24 (cubit fetch/restore wiring, pinned-time behavior, 320dp no-overflow, collapse semantics). CI does not run tests.

## Review asks resolved in-review (`472b34b4`, `0585bea1`)

Two independent fresh-review passes; the second mutation-tested the fixes (reverting each one fails exactly its pinning test).

1. **Pinned-time deploys got blank alternative rows** (trufi-app hides synthetic clock times): same-route groups no longer offer expansion there; visible rows always state their ride.
2. **`maxItineraries` 5→15 reverted to opt-in**: the global bump bought OTP2 deploys ~3× payload without better grouping.
3. **Spacer+Flexible layout regression from the first fix round**: departure times truncated to ~⅓ width on ordinary phones; now `Expanded`+`FittedBox` (scale-down beats mid-word ellipsis for clock times), badge right-aligned via `Expanded`+`Align`.
4. **Pre-existing on `main`, fixed in passing**: the card header overflowed on 12-hour locales at any ordinary phone width — the grouped list now renders a 320 dp phone with zero RenderFlex overflows (pinned by test).
5. **Dead chevron**: collapsing a group held open by its selected alternative now moves the selection back to the representative and really closes it.
6. Selection visibility, grouped-row count in "N routes found", nameless-route feeds, walk-vs-bike zero-slot groups, misleading live badge on multi-route chips, deterministic ordering — all addressed.

## Runtime verification (Android emulator, fully serverless: local planner + OTP 1.5 on localhost over the bundled GTFS)

| Grouped card `P → 106 / 120`, badge `+1 opciones` | Expanded: `Otras opciones`, row `P → 120` with its own times |
|---|---|
|  |  |

| Same-route group: `123 → 120`, `+1 más` | Expanded: `Otras salidas` | Local planner: 3 combos correctly NOT grouped (transfer corners >300 m apart) |
|---|---|---|
|  |  |  |

OTP 1.5 returned 15 itineraries for Valle Hermoso → Matadero Tiquipaya; the list collapses the interchangeable ones ("106 / 120" chip visible in the screenshot) while OTP 2.8 (server-side filtered) behaves exactly as before. Zero exceptions in logcat.

## Follow-ups (non-blocking)

- Pre-existing `RangeError` reachable via a shared deep link with a negative `route` index (`route_planner_cubit.dart`, the guard doesn't cover negatives) — separate issue.
- Grouping tolerances are Cochabamba-calibrated; revisit if a denser-stop city over-groups (they are parameters, no API change needed).
- Edge case: a feed with distinct gtfsIds but no route short names could expand to unlabeled rows under pinned time — unreal for current deploys.
- Pinned-time UX (trufi-app) is covered by widget tests; do an emulator sanity pass when bumping trufi-app.

SamuelRioTz:
Third fresh-review pass audited the product-feedback commit (f0920416): **no blockers** — the main-bus semantics, the option switcher, the vehicle cap and the chip cap all verified clean, including mutation checks (removing the cubit filter, re-widening the grouper, or unwiring the switcher each fails exactly one pinning test). Five minor findings, addressed in f398c845:

1. **Docs drift**: the model header and the cubit comment still described the v1 per-slot semantics ("presentation-only" was no longer true — the vehicle cap rewrites and persists `plan.itineraries`). Rewritten.
2. **Legacy restore**: a plan saved by an older version could restore a selection the vehicle cap has since dropped — map drawing a chain the list no longer shows. `initialize()` now falls back to the best remaining row.
3. **Transit-aware fail-open**: dropping every transit option while a walk-only row survived left "walk 4 km" as the sole result where two chains existed. The cap now reopens whenever no transit option survives.
4. New test pinning that direct and transfer never merge even with the same main bus (the slot-count guard had no dedicated test).
5. Stale v1-semantics comments in tests cleaned up.

Noted, not fixed (inherent): `?route=N` share links index the live provider response, so links generated pre-cap (or by an older web deploy) may select a different row — indices over live OTP responses were already time-sensitive.

Full suite green locally (routing 145, home screen 27; `melos ci:test` + `ci:analyze`).

**Live evidence of the product feedback** (OTP 1.5 local, same OD as before): the vehicle cap turns 15 raw itineraries into 6 rows; the grouped card reads `P → 106 / 120` with `+1 opciones`; the detail shows the option chips and switching re-renders header, timeline and map selection:

| Detail with option switcher | Switched to `P → 120` |
|---|---|
|  |  |

SamuelRioTz:
Second product-feedback round from the maintainer, implemented in 49b58067 + ab129993:

- **Multi-route chips paint one segment per option, each in its own route color**, separated by a drawn divider (contrast-derived — a fixed white bar vanished between light liveries). No "/" text, no "+n" cap: the summary row scrolls.
- **The "+N options" badge and inline expansion are gone** (with their four l10n keys). The group's card shows every route and wears the currently chosen option; the options live in the **detail**, whose switcher now uses the app's visual language: selectable pills with per-route colored badges and the departure time.
- **Two real grouping misses caught live** against the local OTP 1.5 while capturing evidence, both pinned by 1-to-1 mutation tests:
  1. GTFS models one line as several route_ids — comparing by route_id split the 110's departures into separate rows. The main bus is now identified by its **rider-facing name** (route-id fallback for unnamed feeds).
  2. **Only the boarding identifies the main bus**: riding the P five blocks farther to catch the 120 put the alightings 529 m apart and split the canonical `P → 106/120` group. The alighting belongs to the connection.

Fourth fresh-review pass over the feedback commit: **no blockers in executable code** — mutation checks confirmed each new behavior fails exactly one test when reverted; the l10n removal was verified consumer-free and gen-idempotent. Its findings (module doc drift, divider contrast, transparent-color parse guards, switcher a11y semantics) are fixed in ab129993. One **product trade-off flagged for the maintainer**: with name-based identity and boarding-only matching, two DIRECTIONS of the same line boarding at the same corner can merge into one row (the detail switcher still shows each option's real stops). Direct consequence of "the 110 is the 110" — awaiting the maintainer's call; note a naive headsign check would re-split the live 110 case.

| Segmented chip `[P][120¦106]` (per-route colors, drawn divider) | Detail switcher: colored pills | Switched to `P › 106` (stats+timeline+map update) |
|---|---|---|
|  |  |  |

Live run (OTP 1.5 local, Valle Hermoso → Tiquipaya): 15 raw itineraries → 7 rows after the vehicle cap and main-bus grouping; the two grouper fixes each removed a duplicate row on screen. Full `melos ci:test` + `ci:analyze` green (routing 147, home screen 25). Zero exceptions in logcat.

SamuelRioTz:
Final feedback rounds from the maintainer, implemented across 49b58067…52dd731a:

**Product/UX (maintainer-driven):**
- **Mirror grouping**: itineraries left alone after the main-bus pass merge when they share the same final connection ("tienen en común el 120") — on the live dataset, 15 raw itineraries now collapse to **3 rows**, the five X→120 combinations becoming one `[C/J/13/15/138] → 120` row. Main-bus groups keep priority; each mirror group holds every main-bus name at most once (enforced per union-find ROOT — a pairwise guard alone was bridgeable by transitivity, caught with an executable counterexample in round 7); unnamed mains stay out (fail closed).
- **Slanted "/" seams** between chip segments (parallelogram clipping; seam width == slant, documented — flush 5 px after round 6 measured the first cut at 8 px reading as separate tags).
- **Option pills**: bus-type icon + name, no time (the time belongs to the header, which updates on switch; it only joins as a tiebreaker when names repeat). Selection is **color alone** — the saturated pill is the choice, the rest sit dimmed at constant size; no ring.
- Text contrast is picked by computed WCAG ratio against the pill's EFFECTIVE fill blended into the strip's real backdrop — a 0.5-luminance threshold kept white text at ~2.3:1 on saturated colors (the real white/black crossover sits near 0.18). Verified numerically in both themes.

**Review rounds 5–9** (each fix mutation-tested): round 6 caught the mirror pass silently reverting the pinned same-name-far-boarding rule (and the diff editing its protecting test); round 7 caught the union-find bridge and the light-theme contrast; round 8 verified all closures 1-to-1 and caught the wrong luminance threshold; round 9 verified the WCAG formula numerically and caught the test oracle judging a translucent color. All fixed and pinned.

| Mirror group, slanted seams | Final pills: type+name, color-only selection | Dark theme |
|---|---|---|
|  |  |  |

Suites: routing 120 unit + home screen 30, full `melos ci:test` + `ci:analyze` green (CI runs no tests). Zero exceptions across all emulator runs. Follow-ups (non-blocking, from round 9): unify the raw 0.5 threshold still used by the common chip and timeline badge with `_bestContrastOn`; black87 compared as pure black costs ≤0.35 ratio in a 0.01-luminance band.

SamuelRioTz:
## Final design update (v5–v7) and review rounds 10–13

The grouped-options UI went through three more product iterations after the last update, all driven by live maintainer feedback:

- **One shared widget** (`SegmentedRouteChip`): the card's grouped slot and the detail's option switcher now render through the same segmented chip — parallelogram segments meeting on thin slanted "/" seams (a custom overlap row decouples seam thickness from the slant angle), with real `Text`/`InkWell` children so tests can find and tap them.
- **Matching order everywhere**: `slotRoutes` now lists options in the same order as `ItineraryGroup.alternatives` (representative first, then departure order), so the card and the detail never disagree (protected by a grouper test).
- **Final selection language**: every segment keeps its route color — the chosen one rides it at full strength, unchosen ones sit slightly muted (82% over the strip backdrop). No ring, no border, no size change. Ink is white whenever it clears WCAG 3:1 on the effective fill, near-black only on too-light fills. The ridden segment announces `selected` to screen readers (pinned by a semantics test).

**Adversarial review rounds 10–13** (fresh reviewer each round, mutation-checked oracles): round 10 found 5 issues (render-object intrinsics, split contrast policy, a11y `selected`, weakened oracle, stale docs) — all fixed; round 11 found 6 more (bit-exact dry layout, last 0.5-luminance threshold in the detail, unnamed-feed `selected`, rendered-ink contrast, lint, doc) — all fixed; round 12 came back **clean** (pixel-probe on seams, `debugCheckIntrinsicSizes`, threshold mutation); round 13 (micro) caught that deleting the ring oracle had unprotected the `selected` announcement — pinned.

Final look:

| Card | Detail |
|---|---|
| card | detail |

Suites: routing 121 unit + home screen 31 widget tests, `flutter analyze` clean (one pre-existing info). Verified on emulator against local OTP 1.5 and 2.8 (`tools/local-otp/`).

PR #987, opened 2026-08-24
Fix: five saved-place icons fell back to the default pin after saving

### LINKED ISSUE #985: [Bug][v5] Certain icons fail to update when editing a favorite place in "Your places"
**Bug description**

When editing a saved place in "Your places" and changing its icon, some icon selections are not applied correctly. Instead of updating to the icon the user selected, the place is saved with the default location/pin icon.

Icons affected by this issue:

- Heart
- Favorite/ribbon
- Coffee cup (cafe)
- Parking (P)
- Gas station/pump

**Severity:** Medium - does not block usage, but creates inconsistency between what is displayed and what is actually saved.
**Priority:** P3

**Test environment**

- Device: Honor Magic7 Lite.
- OS: Android 16 (MagicOS 10.0).
- App Version: v5.3.0 (Cochabamba).

**Steps to Reproduce**

1. Open the app.
2. Open the side menu.
3. Select "Your places".
4. Tap the three-dot menu on a saved place.
5. Select "Edit place".
6. Change the icon to one of the following: Heart, Favorite/ribbon, Coffee cup, Parking, Gas station.
7. Tap "Save".
8. Observe the place in the list.

**Evidence**

Image

https://github.com/user-attachments/assets/db16984d-c309-4ec5-9d4f-df15762c5f39

**Current Result**

The place is saved and displayed with the default location pin icon instead of the icon the user actually selected. Other icons apply correctly.

**Expected result**

After tapping "Save", the favorite place should update in the list showing the new icon correctly, regardless of which icon was selected.

SamuelRioTz:
Closes #985

## Problem

Editing a place in "Your places" and picking **Heart, Bookmark/ribbon, Coffee, Parking or Gas station** saved fine but rendered back as the generic pin. The icon preview inside the dialog looked correct, which made it look like a save bug — it is not: the stored `iconName` was always right.

## Root cause

Two divergent sources of truth, both private, born in the same commit and never synchronized:

- `EditPlaceDialog._iconOptions` offers **16** keys.
- `SavedPlaceTile._getIconFromName` only knew **10** of them; the missing five (`favorite`, `bookmark`, `cafe`, `parking`, `gas`) fell into `default: Icons.place_rounded`.

Exact 1:1 match with the five icons reported in the issue. A sixth latent divergence: `shopping` showed a **bag** in the selector but a **cart** in the list.

## Fix

- New shared table `savedPlaceIcons` (`saved_place_icons.dart`): one entry per key with its outlined (selector) and rounded (list) variants, plus resolver `savedPlaceRoundedIcon()` with the pin as explicit fallback.
- The dialog and the tile now both consume that table; the private duplicate list and switch are gone.
- `shopping` unified to the bag variant on both surfaces (what the user picks is what they see).
- **No data migration needed**: storage was always correct, so places already saved with the affected icons render correctly as soon as this ships.

## Tests

New `saved_place_icons_test.dart` (6 tests, all failing on `main` for the affected keys):

- every selectable key resolves to its own rounded variant (mutation-checked: dropping any key kills 3 tests);
- no key other than `place` falls back to the pin; keys are unique;
- unknown/null keys fall back to the pin;
- **widget tests**: `SavedPlaceTile` renders the exact icon for each of the 16 keys, and the pin for a legacy/unknown key.

Package suite: 24/24 green. `flutter analyze`: clean.

## Follow-ups (out of scope, will file separately)

- The search screen ("Your places" inside the location search) drops `iconName` entirely when converting to `SearchLocation` — no custom icon ever shows there, including the 11 that worked. Deserves its own issue.
- `SavedPlace.copyWith` cannot clear `iconName` (uses `??`), harmless today but a latent trap.

### DIFF
--- a/packages/screens/trufi_core_saved_places/lib/src/models/saved_place_icons.dart
+++ b/packages/screens/trufi_core_saved_places/lib/src/models/saved_place_icons.dart
@@ -1,0 +1,80 @@
+import 'package:flutter/material.dart';
+
+/// Single source of truth for the icons a custom saved place can carry.
+///
+/// The edit dialog offers every entry of [savedPlaceIcons] and the list tile
+/// resolves the persisted `iconName` through [savedPlaceRoundedIcon]. Keeping
+/// both surfaces on this one table is what guarantees that every icon a user
+/// can pick also renders back after saving (#985 — the selector and the tile
+/// used to keep separate lists, and the five keys missing from the tile fell
+/// back to the generic pin).
+class SavedPlaceIcon {
+  const SavedPlaceIcon(this.name, this.outlined, this.rounded);
+
+  /// Stable key persisted in `SavedPlace.iconName`. Never rename existing
+  /// keys: they live in the user's storage.
+  final String name;
+
+  /// Variant shown in the edit dialog selector.
+  final IconData outlined;
+
+  /// Variant shown in lists and tiles.
+  final IconData rounded;
+}
+
+/// Every icon offered by the edit dialog, in display order.
+const List<SavedPlaceIcon> savedPlaceIcons = [
+  SavedPlaceIcon('place', Icons.place_outlined, Icons.place_rounded),
+  SavedPlaceIcon('star', Icons.star_outline, Icons.star_rounded),
+  SavedPlaceIcon('favorite', Icons.favorite_outline, Icons.favorite_rounded),
+  SavedPlaceIcon('bookmark', Icons.bookmark_outline, Icons.bookmark_rounded),
+  SavedPlaceIcon('school', Icons.school_outlined, Icons.school_rounded),
+  SavedPlaceIcon(
+    'shopping',
+    Icons.shopping_bag_outlined,
+    Icons.shopping_bag_rounded,
+  ),
+  SavedPlaceIcon(
+    'restaurant',
+    Icons.restaurant_outlined,
+    Icons.restaurant_rounded,
+  ),
+  SavedPlaceIcon('cafe', Icons.coffee_outlined, Icons.coffee_rounded),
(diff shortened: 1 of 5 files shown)


SamuelRioTz:
## Review — verified on emulator, both ways

**Verdict: works as described.** Root cause confirmed 1:1 — the dialog's 16-key list vs the tile's 10-case switch; the five missing keys are exactly the five icons reported in #985.

- Code: single shared table for both surfaces, explicit pin fallback, no data migration needed (storage was always correct). `flutter analyze` clean.
- Tests: 25/25 green locally (note: CI does not run tests). Mutation-checked both ways: dropping a key from the resolver kills 3 tests; deleting a whole table row kills the key-set pin test added after the fresh-eyes audit (`60475df`).
- Fresh-eyes audit of the fix commit: table is faithful 16/16 to the original selector, the 9 previously-working rounded icons are unchanged, no leftover consumers of the old switch anywhere in the monorepo. Only visual change for existing data besides the five fixed icons: `shopping` now shows the bag (what the selector always offered) instead of the cart.

### Evidence (Android emulator, apps/example, Cochabamba assets)

| Before (main) — coffee picked | Before (main) — list shows pin | After (fix, same data) — coffee | After — edited to heart |
|---|---|---|---|
|  |  |  |  |

The third capture is the interesting one: the place was saved with the coffee icon **under main** (showing the pin), then the fix build was installed **without clearing data** — the same stored place now renders the coffee cup. No migration, as claimed.

1. The location-search screen drops `iconName` when converting saved places to `SearchLocation` — custom icons never show there (pre-existing, also on main).
2. `SavedPlace.copyWith` cannot clear `iconName` (`??` semantics) — latent trap, harmless today.

Issue #986, opened 2026-08-14
[Bug][v5] Trufi Planner suggests a significantly different number of routes between Web and Mobile for the same origin/destination


MaiteAraucoPorrez:
**Bug description**

When searching for routes using Trufi Planner for the same origin and destination, the number of suggested routes varies significantly between platforms: web returns less than 10 routes while mobile returns more than 10 routes.

Additional note: Both, web and mobile routes are being rendered as a straight line between origin and destination instead of following the streets (routing failure).

**Severity:** Low - does not block usage, but produces inconsistent results between platforms for the same search.
**Priority:** P4

**Test environment**

- Device: Honor Magic7 Lite.
- OS: Android 16 (MagicOS 10.0).
- App Version: v5.3.0 (Cochabamba).

**Steps to Reproduce**

1. Open the app (Web and Mobile).
2. Make sure the selected routing engine is Trufi Planner.
3. Set the same origin and destination on both platforms (e.g. "Parque Jorge Trigo Andia" → "Sombrero de Chola, Ciclovía Norte").
4. Search for routes.
5. Compare the number of suggested routes between web and mobile.

**Evidence**

**Web:**

<img width="1581" height="963" alt="Image" src="https://github.com/user-attachments/assets/53df24eb-c3fb-42d9-b076-660638dc4c28" />

**Mobile:**

<img width="612" height="1350" alt="Image" src="https://github.com/user-attachments/assets/d344c952-98ec-4289-bd81-ac409ccff30f" />

**Current Result**

On Mobile, the top route is intermittently rendered as a straight line rather than following the streets. This has occurred multiple times, including after previously appearing resolved following a reinstall. Additionally, mobile consistently suggests significantly more routes than web for the same search.

**Expected result**

Routes should always be rendered following the actual street network, consistently across app sessions and both platforms should also suggest a consistent (or at least similar) number of routes for the same origin and destination.

SamuelRioTz:
Both halves of this issue are now deployed.

**Route count (web < 10 vs mobile > 10).** The web planner's server hard-coded `maxResults: 5` — the same cap behind #929/#952. trufi-server-planner#4 (honor the client's `maxResults`, default 20, clamped 1–20) is merged and **deployed to planner.trufi.app** today. Verified from outside: `POST /api/plan` for a dense Cochabamba pair returned **5 paths before, 20 after** (and exactly 5 when the client asks for 5). Web and mobile now use the same limits, so the counts line up.

[...]

PR #988, opened 2026-08-24
Fix: transfer legs lost geometry at repeated stops on circular lines

SamuelRioTz:
Addresses the client-side part of #986 (the intermittent "route renders as a straight line" report).

## Problem

With Trufi Planner as the routing engine, an itinerary's bus leg sometimes rendered as a straight line between origin and destination instead of following the streets. It only happened on itineraries **with a transfer**, and only for some origin/destination pairs — hence the "intermittent" feel.

## Root cause

Transfer candidates are enumerated from precomputed `PatternConnection`s, whose `myStopIdx`/`otherStopIdx` point at the **actual visit** of the shared stop within each pattern. Those indices are used to score and build the segment — but not stored. Later, `_resolvePathStops` (which attaches stops and shape geometry to the final top-N results) **re-derived** the board/alight positions from the stop IDs via `RoutePattern.indexOfStop`, which returns the **first occurrence** of a stop on the pattern.

On circular lines that visit a stop twice (common in the Cochabamba feed), a leg that boards after the shared stop's first visit re-derives to `toIdx <= fromIdx`, fails the sanity check, and is returned **with no stops and no shape**. The map layer skips legs with empty geometry, so only the two straight 2-point walk legs get drawn: a straight line from origin to destination.

Direct (0-transfer) itineraries were never affected because both endpoints are found with `indexOfStop` consistently.

## Fix

`RoutingSegment` now carries `fromIdx`/`toIdx` recorded at build time, and `_resolvePathStops` prefers them; the ID-based derivation remains only as a fallback for segments deserialized from JSON (which already come with resolved stops). No serialization changes — the remote planner server resolves geometry before serializing.

## Tests

New `repeated_stop_geometry_test.dart`: a synthetic circular feed (`S1 → X → S3 → S4 → X → S6` plus a transfer line at `X`) where the **only** possible itinerary transfers at the repeated stop's second visit:

- the transfer itinerary exists;
- every transit leg carries stops and shape geometry (fails on `main`: the L leg comes back empty);
- the L leg rides `S3 → S4 → X` — the second visit, not backwards.

Reverting the fix kills both geometry tests. Package suite: 18/18, `dart analyze` clean. `trufi_core_routing` suite: same result as `main` (the only failures on both are the pre-existing network-dependent OTP integration tests; CI runs no tests, verified locally).

## Not in scope (same issue, different causes)

- **Web showing fewer routes than mobile**: the deployed web planner server hardcodes `maxResults: 5` — that is trufi-server-planner#4 (same root cause as #929), a deploy, not core code.
- **Walk-only row as a straight line**: for pairs ≤ 1.5 km the planner inserts a walk-only itinerary whose polyline is by design a 2-point line (there is no pedestrian street routing in Trufi Planner). Behaves identically on web and mobile.

### DIFF
--- a/packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart
+++ b/packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart
@@ -26,6 +26,14 @@ class RoutingSegment {
   /// Computed in O(1) via the pattern's precomputed `cumDist`.
   final double transitDistance;
 
+  /// Board/alight positions in `pattern.stopIds`, recorded at build time.
+  /// `-1` when unknown (e.g. segments deserialized from JSON). They must be
+  /// carried instead of re-derived from the stop IDs: on patterns that visit
+  /// a stop twice (circular lines) `indexOfStop` returns the FIRST occurrence,
+  /// which is not necessarily the one this segment rides through (#986).
+  final int fromIdx;
+  final int toIdx;
+
   const RoutingSegment({
     required this.route,
     required this.fromStop,
@@ -36,6 +44,8 @@ class RoutingSegment {
     this.scheduledDuration,
     this.shapePoints = const [],
     this.transitDistance = 0,
+    this.fromIdx = -1,
+    this.toIdx = -1,
   });
 
   /// Stop IDs for this segment (derived from stops).
@@ -463,6 +473,8 @@ class GtfsRoutingService {
       pattern: pattern,
       transitDistance: transit,
       scheduledDuration: estDuration,
+      fromIdx: fromIdx,
+      toIdx: toIdx,
     );
   }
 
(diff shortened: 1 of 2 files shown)


SamuelRioTz:
## Review — root cause verified against the real feed, both ways

**Verdict: correct and surgical.** Fresh-eyes audit of the fix commit came back clean, with one sharpening: the ONLY build-time index in the whole pipeline that can point at a non-first visit is `conn.myStopIdx` (→ the first leg's `toIdx`); `otherStopIdx` and every direct-route index always come from `indexOfStop`. This fix covers exactly that case and changes nothing else.

**The bug is live in the production feed.** Sweeping the embedded Cochabamba GTFS: 3 patterns visit a stop twice, and there is a concrete reproducing pair — *Calle Batalla de Calama & Hamiraya → Calle Manuel Virreira & Av. Fuerza Aérea* (300 m walk): the route-52 transfer itinerary ranks **1st of 5**, boards at index 36 / alights at 43, while `main`'s re-derivation yields **36→27** — the leg comes back with no stops and no shape, and the map draws a straight line.

- Tests: 3 new (transfer exists / every leg keeps stops+geometry / the leg rides the SECOND visit, `S3→S4→X`). Reverting the fix kills 2 of them. Package 18/18, analyze clean.
- Consumer suite `trufi_core_routing`: non-network failure set identical to `main` (only pre-existing OTP integration tests fail on both; `--exclude-tags integration`: 125/125 on both). Note: CI runs no tests — all run locally.
- No serialization change, deliberately: the remote planner server resolves geometry before serializing, and `fromJson` rebuilds `pattern` as a stub the original indices wouldn't fit. Old server + new client degrades to exactly `main`'s behavior.

### Emulator (common-case parity, Android, apps/example)

Same user pair on `main` and on this branch: **15 routes found on both**, top itinerary follows the streets. The broken case isn't reachable in the example UI (fixed 800 m walk suppresses the transfer bucket for this feed) — it's pinned by the unit tests and the feed sweep above.

| main — 15 routes, streets | fix — 15 routes, streets |
|---|---|
|  |  |

### Release note (not a merge blocker)

This package also runs inside **trufi-server-planner**: the web planner won't get the fix until the server is rebuilt against the fixed core — best done together with deploying trufi-server-planner#4 (the hardcoded `maxResults: 5`), which kills the other half of #986 (web <10 vs mobile >10 routes) in the same redeploy.

### Follow-up (separate issue)

Pre-existing, found during the audit: pattern connections are only recorded at the **first** visit of the shared stop on the *destination* pattern — circular destination patterns can miss valid transfer candidates. Completeness, not geometry; untouched by this PR.

Issue #984, opened 2026-08-14
[Bug][v5-Cochabamba] Mobile origin and destination search returns limited place suggestions


jheimc:
## 📝 Description
The origin and destination search on mobile returns limited suggestions when searching for common place names or locations.

For example, when searching for **"universidad"**, the mobile app only returns streets or addresses that contain or are associated with the searched word, instead of also suggesting relevant places such as universities.

This behavior differs from the web version, where the same search returns a broader list of suggestions and includes relevant places categorized as locations, such as different universities.

The same issue can be reproduced with other searches such as **"quillacollo"**, **"correo"**, and similar location-related terms.

**Severity:** Medium – Limits the ability to discover and select relevant places through the origin/destination search.  
**Priority:** Medium – Affects the search experience on mobile but does not completely prevent users from entering an origin or destination.

## 📱 Test Environment
- **Device:** Mobile: Redmi Note 11 Pro+
- **OS:** Android 13
- **Application:** Trufi App v5.3.0
- **Platform:** Mobile
- **Comparison:** Trufi Web

## 🛠 Steps to Reproduce
1. Open Trufi App v5.3.0 on a mobile device.
2. Open the route planner.
3. Select the **Origin** or **Destination** search field.
4. Enter **"universidad"**.
5. Review the suggested results.
6. Repeat the search using terms such as **"quillacollo"** or **"correo"**.
7. Perform the same searches on the web version of Trufi.
8. Compare the suggestions returned by mobile and web.

## 📎 Evidence

**Mobile:** Search results only show streets or addresses related to the searched term.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/932f15b2-507e-4707-b7d2-1da82a1f2f76" />

**Web:** Search results include a broader list of relevant places, including locations such as universities.

<img width="1904" height="960" alt="Image" src="https://github.com/user-attachments/assets/d8480071-e9b5-4c9a-ab83-d86e6208ce83" />

## ❌ Actual Behavior
On mobile, searches such as **"universidad"**, **"quillacollo"**, and **"correo"** return a limited set of suggestions, primarily streets or addresses containing or associated with the searched term.

Relevant places that match the search term are not displayed as suggestions.

## ✅ Expected Behavior
The mobile origin and destination search should return relevant places in addition to streets and addresses, providing a similar range of suggestions to the web version.

[...]

PR #989, opened 2026-08-24
Search: interleave sources and merge near-duplicates so places surface next to streets

SamuelRioTz:
Addresses the client-side part of #984 (mobile search showing "only streets").

## What actually happened (two causes)

**Primary — infrastructure, already fixed:** the issue was filed on 2026-08-14 at 03:13 UTC from an Android 13 device — ~12 hours **before** the Cochabamba TLS certificate was re-issued that same day (the Let's Encrypt Gen-Y chain broke HTTPS on Android ≤ 13; same incident as Lima). With Photon unreachable, the composite search silently swallowed the failure and only the offline street index answered: "universidad" → 1 street, "correo" → 0 rows. Photon itself was never filtering anything — today the exact query the app sends returns 10 universities. This PR cannot fix that (it is a certificate-chain rule, documented in ops), but it fixes what made the degradation invisible and what still looks wrong with Photon healthy:

**Secondary — real UX problem in core, this PR:**
1. Results were concatenated **source by source**: up to ten offline street rows always ranked above the first online place. On a phone screen (~5-6 visible rows) that reads as "the app only finds streets" even with Photon healthy — searching "quillacollo" shows 10 street rows before the first place.
2. Geocoders return **near-copies** of the same POI or street a few blocks apart (Universidad Latinoamericana ×3, Calle Manuel Virreira ×4), eating the 10-result budget.

## Fix

- **Round-robin interleave** across services (first result of each, then second of each, …): streets and places share the first screen. Ties keep the `services` order, so offline junctions still lead.
- **Near-duplicate merge**: same display name (case/space-insensitive) within `nearDuplicateEpsilon` (default ~220 m, configurable) collapses into the first hit. A nice side effect: an online pin for a street collapses into the richer offline row that carries the corner drill-down.

No changes to per-service limits, timeouts, failure handling, drill-down, language forwarding or `reverse`.

## Tests

3 new unit tests (interleave order, near-copies merge, online-pin-collapses-into-offline-street), plus the existing 7 composite tests unchanged — including "same name far apart both stay" (different places named "Mercado" 3–5 km apart survive the merge). Mutation-checked: disabling the near-merge kills 2 tests; reverting to concatenation kills the interleave test. Package unit+widget: **130/130**; `flutter analyze` clean. (The 10 Photon *integration* tests hit the live network and fail identically on `main` in this environment; CI runs no tests.)

## Follow-ups (separate issues, out of scope)

- Silent degradation: the composite swallows a failing source (`catch → []`) with no signal to the UI — the TLS outage looked like "search is broken". Deserves a degradation flag + a hint in the search screen.
- Offline POIs: `OfflineSearchDataService` never parses the `pois` key (and trufi-app's slimmed index ships it empty), so there is no place search without network.
- Ops: monitor that every city cert chain keeps ending in ISRG Root X1 (the pending-improvements list already tracks this).

### DIFF
--- a/packages/trufi_core_search_locations/lib/src/services/composite_search_location_service.dart
+++ b/packages/trufi_core_search_locations/lib/src/services/composite_search_location_service.dart
@@ -15,9 +15,17 @@ import 'search_location_service.dart';
 /// - services are queried **concurrently**; a service that fails or times
 ///   out is skipped, so losing the network degrades to offline results
 ///   instead of an error screen;
-/// - results keep the order of [services] (put the offline one first if
-///   junctions should lead), deduplicated by coordinates so the same
-///   place found twice appears once;
+/// - results are **interleaved** round-robin across [services] (first
+///   result of each, then second of each, …) so no source can push the
+///   others below the fold — with streets offline and places online, the
+///   old source-by-source concatenation buried every place under up to
+///   ten street rows on a phone screen (#984). Ties inside a round keep
+///   the order of [services] (put the offline one first if junctions
+///   should lead);
+/// - duplicates are merged: same id, or same display name within
+///   [nearDuplicateEpsilon] — geocoders return the same POI or street
+///   several times a few blocks apart, and those near-copies used to eat
+///   the result budget (#984);
 /// - [reverse] returns the first non-null answer, in order.
 class CompositeSearchLocationService
     with SearchLocationDrillDown, LanguageAwareSearch
@@ -31,10 +39,15 @@ class CompositeSearchLocationService
   /// the same place when merging.
   final double dedupeEpsilon;
 
+  /// Results with the same display name closer than this (in degrees,
+  /// ~2e-3 ≈ 220 m) are treated as near-copies of one place and merged.
+  final double nearDuplicateEpsilon;
+
   CompositeSearchLocationService({
     required this.services,
     this.timeout = const Duration(seconds: 8),
     this.dedupeEpsilon = 1e-5,
+    this.nearDuplicateEpsilon = 2e-3,
   }) : assert(services.isNotEmpty, 'at least one service is required');
 
   @override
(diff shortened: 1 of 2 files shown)


SamuelRioTz:
## Review — verified on emulator with a healthy geocoder, both ways

**Verdict: works, and the fresh-eyes audit earned its keep.** The second-pass review of the first commit found a real regression the interleave introduced — a drillable offline street arriving in a later round could be swallowed by a plain online pin for the same place (first-arrival-wins), silently losing the corner drill-down. Fixed in `e9f005ed`: near-duplicates now **replace in place** when the newcomer is drillable and the incumbent is not, with a test pinning the exact cross-round scenario. Same commit adds a radius-edge test (same name ~450 m apart stays two results — the old far-apart guard only broke past ~5.5 km) and syncs the `dedupeEpsilon` docs (legacy floor, `max()` semantics).

- Tests: **132/132** unit+widget, analyze clean. Mutation-checked all three mechanisms: near-merge off kills 2, reverting to concatenation kills 1, dropping the in-place replacement kills 1. The 10 Photon *integration* tests hit the live network and fail identically on `main` (CI runs no tests — all local).
- Consumers audited: the search screen renders per-item and resolves the fork icon via `canDrillDown` per element — nothing assumes "streets first" or indexes by position. Only in-repo consumer is apps/example (offline+photon, the designed case).

### Evidence (Android emulator, apps/example, live photon.trufi.app)

| "universidad" on main | "universidad" fixed | "quillacollo" on main | "quillacollo" fixed |
|---|---|---|---|
|  |  |  |  |

Before: *Universidad Latinoamericana* ×3, *Cosmos* ×2, *Calle Quillacollo* ×2. After: all unique. Known limit, visible in the last capture: the three bare "Quillacollo" rows survive — they are the municipality as different OSM elements **more than 220 m apart**; collapsing those needs place-type-aware rules (follow-up, not this PR).

### Context on the original report

The issue was filed from an Android 13 device ~12 h **before** the Cochabamba TLS certificate re-issue on 2026-08-14 (Let's Encrypt Gen-Y chain broke HTTPS on Android ≤ 13). With Photon unreachable, the composite silently degraded to offline streets only — that was the "no universities at all" part, and it is already fixed server-side. This PR fixes what still looked wrong with Photon healthy (burying and duplicates) and is also what made the outage read as "search only finds streets".

PR #990, opened 2026-08-24
Saved places: duplicate guard at human scale, renamed Home/Work shows its name (#898)

SamuelRioTz:
Addresses the two reports on the reopened #898 (v5.3.0). Both live in `trufi_core_saved_places`.

## 1. Duplicates were still accepted

**Why the fix from #956 did not hold.** The guard compared coordinates with a `1e-6°` epsilon (~11 cm). The map picker returns the raw camera centre, and one screen pixel at zoom 15 is ~4.5 m — about 40× the epsilon — so two taps on "the same spot" never matched. It only fired when the map was not moved at all, which is exactly what the emulator repro did (coordinates differing at 1e-14). In the reporter's hands it never triggered.

**Fix.**
- Same name within **50 m** counts as the same place (`SavedPlacesCubit.duplicateRadiusMeters`, haversine). Names compare trimmed, case- and accent-insensitive, with combining marks stripped ("Café" == "cafe", also as NFD).
- The guard moved from the two screen copies into the cubit's single write path: `savePlace` / `updatePlace` now return `bool` and persist nothing on a duplicate; the screens only show the existing `placeAlreadySaved` snackbar. No new l10n keys.
- Edits that keep a place's own identity (icon, type, a nudge inside the radius) are always accepted, so entries duplicated before the guard existed stay editable (the #956 fresh-review finding is preserved). History entries are never checked.
- Semantics kept from #956: same name **far** away is a different place (two "Farmacia" in different neighbourhoods); a different name at the same spot is allowed.

## 2. Renaming Home/Work did not show in the list

`SavedPlaceTile._getDisplayName` replaced the stored name with the localized "Home"/"Work" for every home/work place whose name had no `_` — a stale heuristic that never matched real data, so the rename was saved (the edit sheet showed it) but the card never did. The label now falls back to the localized default **only** when the stored name is empty or is itself a default label in any supported language (`Home`/`Casa`/`Zuhause`, `Work`/`Trabajo`/`Arbeit`), so existing entries keep following the UI language and a renamed Home shows the user's text. The helper `savedPlaceDisplayName()` is exported for other surfaces.

## Tests (44/44 locally — CI does not run tests)

- `duplicate_places_test.dart` (20): haversine sanity, name normalisation (incl. NFD), the reporter's case (2 m off → duplicate), 10 m / 500 m radius edges, home/work count, history never counts, `savePlace` rejects and persists nothing, `updatePlace` keeps pre-existing duplicates editable and rejects renaming onto another place.
- `default_place_label_test.dart` (6): renamed Home/Work shows the name; default label stays localized and follows the language; empty name falls back; favourites unaffected.
- `saved_places_screen_duplicate_test.dart`: drives the real `SavedPlacesScreen` — "+" → name → pick on map → Save, twice with the picker 2 m apart → one entry + the warning; a third pick 500 m away is a new entry.

Mutation-checked: radius back to 10 cm kills 4 tests (incl. the screen flow); the old label logic kills 3.

## Design decisions worth knowing (surfaced by the fresh-eyes audit)

- **Radius trade-off.** Two genuinely different places with the same name closer than 50 m (two "Farmacia" on one corner) can no longer both be saved under that name — rename one. The alternative the issue title suggests (block by name only, anywhere) is a one-line change if preferred.
- **Setting Home/Work for the first time on top of an existing favourite with the same name (within 50 m) is rejected**, same as under #956. To "promote" a favourite to Home, edit the favourite and change its type (identity kept → accepted).
- **A Home renamed to exactly a default label of another language ("Home" while in Spanish) keeps following the UI language.** Price of keeping existing default entries localized; favourites always show their own name.

### DIFF
--- a/packages/screens/trufi_core_saved_places/lib/src/cubit/saved_places_cubit.dart
+++ b/packages/screens/trufi_core_saved_places/lib/src/cubit/saved_places_cubit.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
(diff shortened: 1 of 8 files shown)


SamuelRioTz:
## Review — reproduced on the real app, verified both ways

**Verdict: works as described.** Both reports on the reopened #898 were reproduced first on **trufi-app 5.3.0** (the build the reporters used) on an Android 13 emulator, then re-run on this branch.

- **Why #956 did not hold**: two hand picks "on the same spot" landed at −17.39951 and −17.39959 (**~9 m** apart) — the 1e-6° (~11 cm) epsilon could never match that. The picker returns the raw camera centre; the old guard only fired when the map was not moved at all, which is what the original emulator repro did.
- Code: same normalized name within **50 m** (haversine) is a duplicate; the check moved into the cubit's single write path (`savePlace`/`updatePlace` return `bool`), the two screen copies just show the existing snackbar. #956 semantics kept: same name far away is a different place, a different name at the same spot is fine, pre-existing duplicates stay editable. History never checked. `flutter analyze` clean, no new l10n keys, no data migration.
- Home/Work label: the tile no longer overrides the stored name for every home/work — it falls back to the localized default only when the name is empty or is itself a default label in any supported language, so existing "Casa"/"Home" entries keep following the UI language and a rename finally shows.
- Tests: **44/44** green locally (CI does not run tests). Mutation-checked: radius back to 10 cm kills 4 tests including the real-screen flow; the old label logic kills 3. Note for anyone writing widget tests on this sheet: on the default 800×600 test surface the Save button is below the fold and `tester.tap` silently misses — the new screen test sets a phone-sized surface.

### Evidence (Android 13 emulator)

| v5.3.0 — two picks 9 m apart | v5.3.0 — **both saved** | v5.3.0 — Home renamed, card still "Casa" | v5.3.0 — …but the edit sheet shows the new name |
|---|---|---|---|
|  |  |  |  |

| Fix — second pick 7 m away | Fix — warning, **one entry** | Fix — same name **720 m** away is allowed | Fix — renamed Home shows in the list |
|---|---|---|---|
|  |  |  |  |

Logcat clean (no unhandled exceptions) across both flows.

1. The location-search screen renders `place.name` raw for Home/Work (`location_search_screen.dart:172-183`); `savedPlaceDisplayName()` is now exported so both surfaces can share the label.
2. `SavedPlace.copyWith` cannot clear `iconName` (`??`) — already noted on #987.
3. The add/edit dialog wiring is still duplicated across two screens (debt from #956); this PR shrank each copy to three lines but did not unify them.

### Fresh-eyes audit of the fix commit

Clean — no functional finding. It re-ran the suite (44/44), `dart format --set-exit-if-changed` and `flutter analyze` on the package, the example and the home-screen consumer, and reproduced both mutation claims. Two text fixes applied in the amended commit (`5c9d410`): the cubit doc comment no longer calls `savePlace` "the single write path" (`addOtherPlace`/`setHome`/`setWork` remain unguarded low-level writers, with no external callers), and the test count in the message. The three design decisions it flagged (radius trade-off, first-time Home over a same-named favourite, default-label-in-another-language quirk) are now spelled out in the PR description.

PR #992, opened 2026-09-09
Planner: walkable transfers and conditional same-name rule (trufi-sanaa#2)

SamuelRioTz:
Fixes the engine side of trufi-association/trufi-sanaa#2 (reopened): the reporter's trip to the Al-Kumaim mall and the reversed university trip answered "no routes" although valid one-transfer itineraries exist. All in `trufi_core_planner`, plus the walk leg and the config knobs in `trufi_core_routing`.

## Root cause — two rules in the transfer index

Both live in the block that precomputes pattern-to-pattern connections (`gtfs_route_index.dart`, formerly L213-253):

1. **Shared `stop_id` only.** A transfer existed only where two patterns served the *same* stop. Sana'a's feed is derived from OSM (`stop_id` = way node), so the two kerbs of a street are different stops: the reversed university trip needs to alight 14 at `504469874` and board 7 at `1223812364`, **15 m** away — never connected. The feed has 1953 stop pairs within 15 m.
2. **Same `route_short_name` ⇒ same line, never transfer.** Written for Cochabamba, where "Línea 209" is split into `route_id` 67/68 by direction. Sana'a is built per OSM relation and 157 of its 194 routes are called "7" — distinct lines, distinct terminals, one informal ref. Every "7 → 7" transfer, i.e. the whole mall trip, was discarded, **even where the two lines share a stop** (the reversed mall trip transfers at `597203155`, a shared stop, and still failed).

Neither #976 (pool size), #978 (k-NN), #983 (grouping) nor #988 (geometry) touched this block; the planner in the reporter's app (v5.22.2) is identical here to `main`.

## Fix

- **Walkable transfers.** `GtfsRouteIndex` takes the `GtfsSpatialIndex` and links a pattern's stop to every other pattern with a stop within `transferRadiusMeters` (default **100 m**, MOTIS' `link_stop_distance`; OTP pre-computes street transfers up to 30 min/2 km). The walk is added to the score with the usual reluctance and surfaces as `RoutingPath.transferWalkDistance`; the second leg boards at the real boarding stop. `0` restores the old shared-stop-only behaviour.
- **Thinned to one connection per crossing.** At a given alight stop only the nearest stop of the other pattern is kept, and along the alight pattern only local minima of the walk. Where two lines run side by side with stops every 30 m every stop is within 100 m of several stops of the other line; keeping all of them multiplied Cochabamba's dense feed from 1.2 M to **16 M** connections for no extra reachability. Thinning keeps every reachable pair (same routeless count with and without it on 340 Cochabamba pairs) but the best score of a pattern pair can move slightly (median +0.1 %, max +14 %); the walk-0 table is untouched. Shared-stop connections are kept exactly as before — a brute-force test pins the walk-0 table to an independent computation.
- **Columnar storage.** Connections live in CSR typed arrays (3×`Int32List` + `Float32List`, 16 B each) behind a read-only `PatternConnections` view (`List<PatternConnection>` for callers, `…At(i)` accessors and `firstIndexAfter()` for the hot loop). Cochabamba's dense feed on `main` holds 1.2 M `PatternConnection` objects (~58 MB); the new table with walkable transfers is 2.6 M entries in **40 MB**. The build reuses one scratch buffer per column and copies each pattern's exact chunk, so its transient memory is one extra copy of the table — not the 7.3 M entries (111 MB) of pre-thinning bounds that an earlier revision pinned through `sublistView` (peak RSS building Cochabamba in the harness: 621 MB now, 774 MB before, `main` 596 MB).
- **Conditional same-name rule.** Each route gets a *line key*: its trimmed `route_short_name` when at most `sameNameRouteLimit` (**3**) `route_id`s carry that name, otherwise its `route_id`. Same key ⇒ same line ⇒ no transfer between them and one itinerary row (`findRoutes` dedupes by line key instead of `shortName|shortName`, so two "7 → 7" answers via different lines are no longer collapsed into one). Cochabamba has no short name shared by more than 3 routes: forcing the rule always-on yields the identical count (2 608 421 = 2 608 421). `TrufiPlannerConfig.local(sameNameRouteLimit:)`, `LocalPlannerClient` and the isolate loader expose the threshold next to `transferRadiusMeters` (a very large value restores the old unconditional rule).
- **Candidate cap 1 500 → 20 000** (`GtfsRoutingService.maxTransferCandidates`). The one-transfer enumeration cap dated from when every candidate built two segments; the search now keeps one path per pattern pair (`bestByPair`), so a candidate is a score comparison and the uncapped search went from 446 ms to 77 ms average on Cochabamba. Meanwhile walkable transfers double the connections per pattern, so at 1 500 the budget ran out on the first origin stops and truncated 24/340 Cochabamba queries to a worse top result than `main` (up to +23 % score). At 20 000 none is worse; see the numbers below.
- **Walk leg in the UI.** `TrufiPlannerProvider.convertToItinerary` inserts a `WALK` leg between two transit legs whose alight and boarding stops differ, with time offsets, so the map no longer draws a jump and the summary counts the walk. `TrufiPlannerConfig.local(transferRadiusMeters:)` exposes the knob; the isolate loader passes it through.
- `GtfsSpatialIndex.findStopsInRadius` is now an exact range query (it silently capped at 100 results) and `forEachStopWithin` visits without allocating.

## Measured (this machine)

| Feed | Connections before → after | Table | Index build before → after |
|---|---|---|---|
| Sana'a (6 179 stops, 194 patterns) | 2 116 → 16 085 | 0.2 MB | 10 → 20 ms |
| Cochabamba, trufi-app feed (23 679 stops, 657 patterns) | 1 207 746 → 2 608 421 (6.3 M with only the nearest-stop-per-pattern rule, 16 M without either thinning rule) | 18 MB → 40 MB columnar (was ~58 MB boxed) | 0.25 → 0.45 s |

Reporter's coordinates through `GtfsRoutingService` with the app's parameters (1500 m, pool 150, 5+5):

| Case | `main` | this branch |
|---|---|---|
| origin → mall Al-Kumaim | 0 | **2** — 7 (18800489) → 63 m walk → 7 (20046785), 8.9 km transit |
| mall → origin | 0 | **3** — first via a 54 m walk, second at the shared stop 597203155 |
| share link → university (worked) | 1 | 2 (same 7 → 14 at the shared stop, plus a second 14) |
| university → share link | 0 | **1** — 14 (18800934) → 15 m walk → 7 (18800748) |

Cochabamba regression, 340 random stop pairs (seeds 42 and 7) at the app's 800 m: 1 178 → 1 241 itineraries, 22 → 18 pairs without a route, 0 itineraries chain two routes with the same short name (the 209 rule holds), and `main`'s shared-stop connection table is reproduced byte for byte (1 207 746 entries, all with walk 0). With the cap at 20 000 the branch never ranks a pair below `main` and never returns a worse top result (at 1 500 it truncated 24/340 queries to a worse top result, up to +23 % score, because twice as many connections per pattern exhaust the budget on the first origin stops). 8.2 ms average per query, 104 ms max (the max is the direct phase; `main` 5.5 ms / 90 ms). Same at 1500 m (0 worse). Sana'a, 344 queries: 182 → 560 itineraries, 203 → 147 without a route, 0 worse.

## Tests

`trufi_core_planner` **56/56** (18 before), `trufi_core_routing` unit **126/126** (122 before). CI does not run tests.

- `sanaa_walkable_transfers_test.dart` (19): a 30 KB cut of the real feed with the ten routes involved; the four coordinate pairs verbatim, ≥1 itinerary each; the specific stops and walks (63 m, shared stop, 15 m); B-fwd unchanged; "7 → 7" rows not collapsed; each rule alone — legacy index reproduces the report exactly (only B-fwd plans), the radius alone rescues B-rev only, the name rule alone rescues A-rev only; the walk-0 table equals the brute-force shared-stop table (`test/support/connection_tables.dart`, 138 entries on the cut, 94 of them right after another shared stop — the entries thinning would remove if it were applied to them); `LocalPlannerClient` passes both knobs; the default candidate cap is a no-op on the cut while a cap that runs out on the first origin stops loses A-rev's best itinerary.
- `transfer_connections_test.dart` (19): two kerbs 15 m apart → one connection with the measured walk, itinerary alights/boards at different stops, radius 0 and radius 10 give none; thinning on a parallel stretch and on two separate crossings; Cochabamba's 209 as `route_id` 67/68 never chained, and the 209/110 corridor keeps a shared-stop connection at every stop (12 = brute force); 3 routes sharing a name still one line, 5 routes named "7" all chained, limit configurable, same `route_id` always one line; sorted columnar layout, `firstIndexAfter`, accessors, read-only view, walk-0 entries = brute force; exact range query vs brute force.
- `trufi_planner_walk_leg_test.dart` (2): WALK leg between legs with stop ids, distance, contiguous times and total walk; none when the stop is shared.
- `trufi_planner_config_test.dart` (3): both knobs in `local`, fixed in `remote`.

## Notes for the reviewer

- Two "7" lines now appear as "7 → 7" in Sana'a's results; they are different lines (the long names carry the terminals). How the row should label them is a product question, not changed here.
- In Sana'a the line-key dedupe can fill the five direct slots with same-named "7" routes (10/344 random queries lose a row label `main` showed; 13 rows are stop-identical duplicates of another row) — a `(fromStop, toStop)` secondary dedupe is a follow-up.
- Lima gets its first transfers between corridors (0 → 8 connections, 31 new itineraries in 340 random pairs, nothing lost).
- Straight-line 100 m can cross a divided avenue; MOTIS accepts the same approximation by default. Not raised to 150 m: Cochabamba's table grows 6 % more for no new routes in the sample.
- Even at 20 000 the cap still truncates some queries relative to an uncapped search (27/340 Cochabamba queries have a better top result without any cap, median +1.2 %); enumerating by score or per pattern pair is a follow-up.


SamuelRioTz:
## Review — reproduced on the real Sana'a app, verified both ways

**Verdict: works as described.** The two failing trips from trufi-association/trufi-sanaa#2 were reproduced first on **trufi-sanaa `main` (core v5.22.2)** on an Android 16 emulator via the app's own `trufiapp://route?…` deep link, then re-run on the same app with this branch's packages.

- **Root cause, confirmed on both `main` and v5.22.2**: transfers were only precomputed between patterns sharing a `stop_id`, and any two routes with the same `route_short_name` counted as one line. In an OSM-derived feed the two kerbs of a street are different stops (1953 stop pairs within 15 m in Sana'a), and 157 of 194 routes are called "7" — so every "7 → 7" transfer died even at a genuinely shared stop, and the reversed university trip (a 15 m kerb-to-kerb transfer) never existed.
- Code: walkable transfers within 100 m (MOTIS' `link_stop_distance`) through the existing spatial index, thinned to one connection per crossing; a *line key* per route (short name only when ≤ 3 route_ids carry it) drives both the "same line" rule and the itinerary dedupe; connections stored columnar in CSR typed arrays behind a read-only `PatternConnections` view; the second leg boards at the real stop and the provider emits the WALK leg between legs. `flutter analyze` clean, `dart format` applied, `getConnectionsFor` still returns a `List<PatternConnection>` so downstream callers (trufi-server-planner, the issue harness) compile unchanged.
- Measured on the feed **trufi-app actually ships** (23.7k stops, 657 patterns — denser than the `apps/example` feed): the naive radius would have meant 16 M connections; with thinning 2.6 M in **40 MB columnar versus ~58 MB of boxed objects for the 1.2 M shared-stop connections `main` builds today**, and the build's transient memory is one extra copy of the table (peak RSS 621 MB vs 596 on `main`). Index build 0.25 → 0.45 s here. The same-name rule is a measured no-op there (identical table with it forced on). 340 random pairs at the app's 800 m: 1 178 → 1 241 itineraries, 22 → 18 without a route, **no pair and no top result ranks below `main`**, 0 same-name chains (the 209 rule holds), `main`'s shared-stop table reproduced byte for byte; 8.2 ms average per query (`main` 5.5), the maximum unchanged (direct phase).
- Tests: planner **56/56** (18 before), routing unit **126/126** — CI does not run tests. Mutation-checked: default radius back to 0 kills 17 tests, name limit to ∞ kills 11, thinning off kills 2, thinning applied to shared stops too kills 2, candidate cap ignored kills 1, unscored transfer walk kills 1, walk leg removed kills 1.

### Evidence (trufi-sanaa app, Android 16 emulator; before = v5.22.2 pin, after = this branch)

| v5.22.2 — mall: **no routes** | v5.22.2 — university reversed: **no routes** | v5.22.2 — university forward works (7 → 14, 57 min, 833 m — the reporter's screenshot) |
|---|---|---|
|  |  |  |

| Fix — mall: **7 → walk → 7**, 42 min, 9.8 km, 912 m | Fix — university reversed: **14 → walk → 7**, 58 min | Fix — university forward **unchanged** (7 → 14, 57 min, 833 m) | Fix — mall detail: **Walk 0 min (62 m)** between the two "7" lines (different long names) |
|---|---|---|---|
|  |  |  |  |

Logcat clean across all runs (no unhandled exceptions). The engine returns 2–3 itineraries for the mall; the UI shows one row because the #983 grouper groups by the main bus name — both are "7". Cochabamba was verified with the planner harness on trufi-app's real feed rather than in the emulator: trufi-app's current toolchain (Flutter 3.47.2 requires AGP ≥ 8.14, the repo pins 8.13 and AGP 9 breaks two plugins) does not produce a debug APK from a clean copy.

1. Sana'a shows "7 → 7" rows for distinct lines; the row could carry the long names (terminals) — product decision.
2. Shared-stop connections are kept in full (1.2 M on Cochabamba's dense feed); applying the same one-per-stretch thinning to them would cut the table to a fraction.
3. Bumping trufi-sanaa to this core needs six `liveVehicle*` strings in the app's own `l10n_ar.dart`.
4. trufi-server-planner should be rebuilt on this core so the web planner gets the same transfers (`transferWalk` is an optional JSON field, old clients ignore it).

### Fresh-eyes audit

A second pass over the first three commits (14 mutations, 340 random pairs per feed diffed against `main`, memory probes) found three minor issues, fixed in `d254678e` and `2579f8a5`:

1. The one-transfer candidate cap was still **1 500**, a figure from when every candidate built two segments; with one path per pattern pair a candidate is a score comparison, and with twice the connections per pattern the budget ran out on the first origin stops — 24/340 Cochabamba queries got a worse top result than `main` (up to +23 % score). Now `GtfsRoutingService.maxTransferCandidates`, default **20 000**: 0 worse, 8.2 ms average.
2. The index build kept `sublistView`s of every pattern's capacity-sized buffer until the final concatenation — 111 MB of transient bounds for a 40 MB table on Cochabamba. Scratch buffers are reused and each chunk is an exact copy; the table is byte for byte the same.
3. No test pinned the shared-stop table (thinning it too passed every test). A brute-force `sharedStopTable()` now has to match the walk-0 entries exactly on three fixtures.

Also: `sameNameRouteLimit` is exposed in `TrufiPlannerConfig.local` next to the radius, and the doc figures quote trufi-app's feed.

Issue #973, opened 2026-08-11
Offline map assets are extracted once and never refreshed on app update


SamuelRioTz:
`OfflineMapLibreEngine` extracts the style, glyph PBFs, sprites and MBTiles from the asset bundle into the app cache directory on first run — and `_copyAsset` returns early when the target file exists (`offline_maplibre_engine.dart:95`). Nothing ties the extracted copy to the app version, so **an app update that ships new offline assets changes nothing for existing users**: they keep rendering the old style/fonts/tiles until they clear app data or reinstall.

Hit in practice on trufi-sanaa#4 (Arabic glyph/style fixes): installing the fixed APK over the old one kept showing the broken map; only `pm clear` refreshed it. Any offline city app shipping a tiles/style/fonts update has the same silent problem.

**Proposed fix**: stamp the extraction with the app `buildNumber` (marker file in `offline_maps/<id>/`); on mismatch, wipe the engine's folder and re-extract. Cheap (one marker read per boot), correct across updates and downgrades, and no content hashing needed.

PR #994, opened 2026-09-09
Planner: persist the built index so cold starts load a snapshot instead of rebuilding (#993)

### LINKED ISSUE #993: Local planner: persist the built GTFS index so cold starts don't rebuild it
## Problem

The offline Trufi Planner parses the bundled GTFS and builds all its indices (`GtfsRouteIndex` with the transfer connections, `GtfsSpatialIndex`, `GtfsScheduleIndex`) **on every cold start** of the app. `TrufiPlannerDataSource.preload()` runs `compute(_loadAndBuildInIsolate, …)` at startup and nothing is persisted; only the parsed result lives in memory for the session. The map appears immediately because the work runs in an isolate, but the planner stays in `loading` until the build finishes, so a user who opens the app and searches right away waits for it — every time, not just after install.

This was cheap enough to ignore until now. With #992 (walkable transfers) the index roughly doubles its connection table on dense feeds, and the build time with it:

| Cochabamba feed shipped by trufi-app (23 679 stops, 657 patterns) | before #992 | after #992 |
|---|---|---|
| Index build, desktop Dart VM | 0.16 s | 0.44 s |
| Connections | 1.2 M | 2.6 M (40 MB columnar) |

On a low-end phone the same build is several times slower (1–3 s is a fair expectation; not yet measured on device — see below). Sana'a adds ~30 ms. Every future feed refresh or engine improvement will push this further.

## Proposal

Persist the **built** index, not just the parsed GTFS, and rebuild only when the input changes:

1. After a successful build, serialize the indices to the app support directory. The transfer table is already columnar typed arrays (CSR layout) after #992, so it can be written and read back as raw bytes; the spatial and schedule indices need a compact binary form (avoid JSON — size and parse time defeat the purpose).
2. Key the blob by a hash of (GTFS asset bytes or its content fingerprint, `trufi_core_planner` version, and the config knobs that shape the index: `transferRadiusMeters`, `sameNameRouteLimit`, `maxWalkingDistance` if it affects the index). Mismatch → rebuild in the isolate and overwrite. Any read error or format-version mismatch → rebuild (never fail the planner because of a stale cache).
3. Load path on cold start: read blob → validate header/hash → hand the ready indices to `LocalPlannerClient.loadFromParsed`. Target: planner `loaded` in well under the time of the current build on device.
4. Keep the current build as the fallback and as the first-run path (first run after install or after a GTFS update still builds once — that is acceptable and matches the asset-extraction behavior fixed in #973/#975).

## Acceptance criteria

- Cold start with a warm cache: planner ready without running `GtfsRouteIndex`'s build; measured on an Android emulator via logs for both trufi-app (Cochabamba) and trufi-sanaa, before/after, with the numbers in the PR.
- Cache invalidation verified both ways: changing the GTFS asset (or the core version / index knobs) triggers a rebuild and the new data is what the planner uses — precedent #973, where an extraction cache survived updates and users kept stale maps.
- Corrupt or truncated blob → silent rebuild, no crash, covered by a test.
- Web unaffected (web uses the remote planner); no new permissions.
- Blob size reported for both feeds; if it is larger than the GTFS zip itself by a wide margin, say so and justify.

## Related

- #992 — walkable transfers (the change that makes the build cost visible).
- trufi-association/trufi-sanaa#2 — the report that led to #992.
- #973 / #975 — the offline asset extraction cache and its invalidation bug (same class of risk).

SamuelRioTz:
Closes #993. The offline planner parsed the bundled GTFS and built every index (`GtfsRouteIndex` with the #992 transfer table, `GtfsSpatialIndex`, `GtfsScheduleIndex`) on **every** cold start, and the planner stayed in `loading` until it finished. The built state is now persisted as a binary snapshot in the app's cache directory and restored on the next start; the GTFS is only parsed and the indices only built when there is no valid snapshot for exactly this feed, these knobs and this engine version. All in `trufi_core_planner` (pure Dart) plus the store and the wiring in `trufi_core_routing`.

## What is rebuilt today, and what it costs

Instrumented run on the desktop VM (JIT), the feeds the apps ship (`harness-output/phases_desktop_jit.txt`):

| Phase | Cochabamba (trufi-app: 23 679 stops, 657 patterns, 104 592 stop_times, 357 597 shape points) | Sana'a (6 179 stops, 194 patterns) |
|---|---|---|
| Parse `shapes.txt` | **0.69–0.78 s** | 31–34 ms |
| Parse `stop_times.txt` | 0.24–0.27 s | 17–21 ms |
| Parse the rest + inflate | 0.13–0.17 s | 26–34 ms |
| `GtfsSpatialIndex` | 52–56 ms | 14–18 ms |
| `GtfsRouteIndex` patterns | 50–71 ms | 4–8 ms |
| `GtfsRouteIndex` connections (#992) | 0.32–0.45 s | 14 ms |
| `GtfsScheduleIndex` | 36–45 ms | 1 ms |
| **Cold start** | **1.58–1.73 s** | 115–120 ms |

Parsing is 68 % of the cold start (`shapes.txt` alone 45 %); the transfer table #992 made visible is 25 %. Persisting only the indices would have left 1.1 s of 1.6 s on the table, so the snapshot holds the **parsed data too**.

## Design

- **`PlannerIndexBundle`** (`trufi_core_planner/lib/src/snapshot/`): the four objects the planner keeps in memory, with `build(gtfsZip)` — the old isolate body — reporting a per-phase timing split.
- **`PlannerIndexCodec`**: `encode(bundle, fingerprint:)` / `decode(bytes, fingerprint:, transferRadiusMeters:, sameNameRouteLimit:)` / `fingerprint(bytes)`. One blob, host byte order, 8-byte-aligned sections with a tag and a length, over a deduplicated string pool written as the **last** section (its offset sits in the header), so the encoder streams the whole blob into a single buffer:
  - `GtfsData` **lossless**: coordinates, `cumDist` and `shape_dist_traveled` as `Float64`, GTFS times as `Int32` seconds, dates as microseconds-since-epoch `Float64` + UTC flag, enums by value, every `Map` in iteration order (the order drives the engine's enumeration order). Integer columns take the narrowest width their maximum fits (1/2/4 bytes) and are widened to `Int32List` on load.
  - Indices as their own working tables: the CSR connection columns of `GtfsRouteIndex` verbatim, the KD-tree of `GtfsSpatialIndex` as its **pre-order** (the tree shape is a function of the stop count — median split — so it is rebuilt in O(n) with no sort and is the same tree), the sorted stop-time groups of `GtfsScheduleIndex`. Each index gains a `restore(...)` constructor; the cheap per-stop / per-route maps are re-derived from the patterns in id order, which inserts every pair in the same order the build did (the round-trip test checks that order for every stop).
  - **Key**: GTFS content fingerprint (two 32-bit FNV-1a streams over 32-bit words + length, 5–7 ms for the 4.9 MB Cochabamba zip, JS-safe arithmetic as in #980), `transferRadiusMeters`, `sameNameRouteLimit`, and `PlannerIndexCodec.formatVersion` — the version of the schema **and** of the index-building rules — and of the parser: the snapshot stores the parsed `GtfsData` too, so any change in `GtfsParser` / `CsvParser` / `*.fromCsv` that yields different data from the same zip must bump it as well (the rule is documented on `formatVersion` and on the parser and the three indices). The guard test digests **every field** of the fixture bundle — parsed data including names, colours, times and shape points; pattern `cumDist`, bbox, headsign, shape; all four connection columns; line keys; per-stop lookup order; KD-tree pre-order; schedule groups — except calendar instants (timezone-dependent, dumped as year-month-day), and pins the digest, so a forgotten bump fails the suite (the #973/#975 failure mode: a cache that survives an update).
  - **Refusal**: wrong magic/trailer, other version, other knobs, other fingerprint, string-pool offset out of range, payload checksum mismatch, truncation, any reference out of range while restoring → `decode` returns `null` with a reason; it never throws for malformed input. Every header bit is validated by its own check (a test flips each one); record counts read ahead of per-record loops are bounded against the remaining bytes before anything is allocated.
- **`TrufiPlannerDataSource.preload()`**: resolves the snapshot path on the root isolate (`path_provider` is a platform channel), then in the worker isolate fingerprints the asset, reads + decodes the snapshot and returns it, or builds as before and also encodes; the root isolate hands the bundle to `LocalPlannerClient` first and writes the snapshot afterwards (async file I/O, unique temp file + rename, the temp file removed on failure, failures only logged). One `debugPrint` line reports source and timings — `planner index loaded from cache in 12ms (read 1ms, decode 8ms; snapshot 1.1 MB, …)` / `planner index built in 1650ms (parse 1091ms, spatial 56ms, index 461ms, schedule 35ms; snapshot 38.2 MB encoded in 369ms; …)` — and `lastIndexLoad` exposes the same for callers and tests.
- **Where**: `<app cache dir>/trufi_planner/<asset>-<hash>.idx`, one file per asset, overwritten when the key differs (no stale accumulation). The **cache** directory rather than application support on purpose: Android Auto Backup caps an app at 25 MB and excludes `getCacheDir()` — a 38 MB file under `getFilesDir()` would make the whole app backup fail (preferences, saved places); iOS keeps `Library/Caches` out of iCloud and may purge it under disk pressure, which is exactly the semantics of a rebuildable cache. Same location as the offline map extraction.
- **Web unaffected**: the store is a conditional import (`planner_index_store_io.dart` / `_stub.dart`), no `dart:io` reaches the web build, `flutter build web --release` of `apps/example` passes. The codec compiles for dart2js (32-bit arithmetic only).
- **Opt-out**: `TrufiPlannerConfig.local(persistIndex: true)` (default); `.remote` has nothing to persist.

## Measured (desktop VM, `harness-output/snapshot_bench_desktop.txt`, `harness-output/fix-0909/`)

| Feed | Build (parse + indices) | Encode (first start only) | Snapshot | Decode | Zip |
|---|---|---|---|---|---|
| Cochabamba (trufi-app) | 1.63–1.70 s | 0.24–0.37 s | **38.25 MB** | **0.17–0.32 s** | 4.92 MB |
| Sana'a | 0.11–0.19 s | 10–57 ms | **1.12 MB** | **8–31 ms** | 0.35 MB |

Re-encoding the restored bundle is **byte-identical** to the original snapshot on both feeds; every field of the restored bundle equals the built one (canonical dump of 3.1 M lines on Cochabamba, 0 differing lines); 300 random stop pairs × 2 walk radii per feed (plus the four trufi-sanaa#2 trips) plan **identical itineraries** from the built and the restored indices (stops, patterns, positions, walks, scores, geometry). Query time is unchanged (Cochabamba 8.3 ms built / 7.8 ms restored).

**Memory on the first start.** The encoder writes into one buffer (string pool last). On the desktop the Cochabamba build alone peaks at 620–628 MB maxRss; build + encode peaks at **687–697 MB** (an earlier draft with a second writer and two copies peaked at 763–775 MB); decoding the snapshot in a fresh process peaks at 444 MB. Numbers from `harness-output/fix-0909/mem_after.txt` (one process per mode).

**Snapshot size vs zip.** Cochabamba's snapshot is 7.8× its GTFS zip. The zip is compressed text and contains no index; 26 MB of the 38 MB are the 2.6 M transfer connections of #992 (6 B each stored — the same table takes 40 MB in memory), 7 MB the 357 597 shape points at full `Float64` precision, the parsed rest ~5 MB (less than the 22 MB of text it replaces). `gzip -1` would bring it to 10 MB (0.22 s to compress, 0.03 s to inflate on the desktop); left as a follow-up because it would trade file size for a decompression pass on every start and blur the measurement here.

## Emulator (trufi-sanaa app, Android 16 arm64 emulator, profile/AOT build)

The real Sana'a app with `dependency_overrides` to `main` (`3ac4838d`) and to this branch; times from logcat — `Preloaded in` is planner-ready measured from the start of `_preloadLocal`, the isolate's own report in parentheses. Full logs and screenshots in the review comment. The "corrupted" and "truncated" rows ran on the variant of the feed without route `18800934` (fingerprint `446118af…`), not on the shipped feed; the conclusion is the same.

| Scenario | Planner ready | Inside the isolate |
|---|---|---|
| `main`, cold start 1 / 2 | **476 / 642 ms** | parse + build, every start |
| branch, first start (no snapshot) | 610 ms | built in 584 ms (parse 311, spatial 33, index 220, schedule 1; **encode 16 ms**) + async write 1.1 MB in 126 ms |
| branch, cold start 2 / 3 | **107 / 81 ms** | loaded from cache in 29 / 43 ms (decode 28 / 39) |
| GTFS without any line "14" (29 route_ids carry that name) installed over it, cache kept | 523 ms | `cache rejected: GTFS fingerprint changed` → built in 496 ms (parse 319, spatial 79, index 20; encode 68); snapshot 1.0 MB (967 012 B, was 1 121 572), snapshot rewritten; the university trip (7 → 14 before) now finds no route on that feed |
| next cold start on the new feed | 72 ms | loaded from cache in 52 ms (decode 50) |
| snapshot corrupted (4 KB overwritten) | 581 ms | `cache rejected: checksum mismatch` → built, rewritten; next start 42 ms from cache |
| snapshot truncated to 300 KB | 463 ms | `cache rejected: truncated snapshot (no trailer)` → built, rewritten |

Debug (JIT) builds keep the ratio but are slower: `main` 1 594 / 1 259 ms; branch 1 516 ms first start (built 1 206 + encode 228), then **598 / 631 ms** from the cache. No unhandled exceptions in any run. Cochabamba could not be run in the emulator (trufi-app's toolchain, same as in #992); it is covered by the harness on the feed trufi-app ships.

## Tests

`trufi_core_planner` **85/85** (56 before), `trufi_core_routing` unit **140/140** (126 before); `flutter analyze` clean in `trufi_core_planner` and `apps/example`, `trufi_core_routing` keeps its two pre-existing infos; `flutter build web --release` of `apps/example` passes.

- `planner_index_codec_test.dart` (29): round trip of every `GtfsData` field (agencies, stops incl. Arabic names, routes, trips, calendars, frequencies, stop times, every shape point), patterns (ids, stops, `cumDist`, bbox, `indexOfStop`, line keys), `getPatternsAtStop` / `getRoutesAtStop` in the same order for every stop, all four connection columns for every entry and the brute-force `sharedStopTable` (#992), the spatial pre-order plus k-NN and range answers at 200 random points, departures at every stop and route frequencies, identical itineraries for the reporter's four trips + 200 random pairs, a `LocalPlannerClient` fed from each; `encode(decode(x)) == x`; decoding from an unaligned view; empty feed; refusal of another fingerprint, other knobs, another format version, wrong magic, garbage, the zip itself, 72 truncation points, 300 random single-bit flips anywhere, every bit of the header, forged record counts (2^31 − 1 agencies / calendars / calendar dates with the checksum recomputed → rejected by the byte bound before anything is allocated) and forged string-pool offsets, trailing bytes and a missing trailer; encode refuses non-whole-second durations; fingerprint sensitivity and known answers; the format-version guard over the canonical dump of every field (`test/support/bundle_dump.dart`).
- `trufi_planner_index_cache_test.dart` (12, routing): real files in a temp cache directory, the fixture served through a mocked asset bundle, the worker isolate as in production — first start builds and writes (temp file renamed, one file per asset), second start loads from cache and plans identically; a changed GTFS asset (fixture minus one route) is detected by fingerprint, rebuilt, snapshot replaced and the new data used; corrupt (flipped byte → checksum), truncated, garbage and other-version snapshots are rebuilt silently and overwritten; other knobs rebuild; `persistIndex: false` never touches the disk; a failing `path_provider` builds without persisting; `clear()` resets the diagnostics and the pending write; **the planner is ready before the snapshot write starts** (the write is held at a test seam between "temp file complete" and "rename": the preload has returned, plans, and no `.idx` exists yet); **`writePlannerIndex` is atomic** (temp file next to the target with the full content while the previous snapshot is still intact, renamed — not copied — over it, stale temp files swept, nothing else left; a failed rename removes the temp file and rethrows).
- `trufi_planner_config_test.dart` (+2): the knob in both modes.

## Notes for the reviewer

- First start after install or update pays build + encode (Cochabamba: +0.24–0.37 s on the desktop) before the planner is ready; the write itself does not hold the planner back (pinned by a test). On the desktop the first start's peak RSS is ~65–75 MB above the build alone for Cochabamba (687–697 vs 620–628 MB maxRss) while the snapshot is encoded into its single buffer. Acceptable by the issue, and identical to the asset-extraction behaviour of #973/#975.
- `GtfsRouteIndex.restore` re-derives `_stopToPatternIds` / `_stopToRoutes` / `_routePatterns` from the patterns instead of persisting them; the equivalence (including insertion order) is argued in its doc comment and checked by the round-trip test on the fixture.
- `GtfsScheduleIndex` is built and now persisted although nothing in the UI reads it yet; dropping it from the snapshot would save ~0.5 MB on Cochabamba — not done, to keep the bundle equal to what the data source exposes.
- Calendar dates are persisted as the local-time instants the parser creates; nothing reads them yet (`getNextDepartures` ignores calendars). When the schedule starts using them, persist year-month-day or add the timezone to the key.
- Follow-ups: gzip the snapshot (numbers above); remove a stale `.idx` when an app renames its GTFS asset (one file per asset name today); overlap `rootBundle.load` and the cache-path lookup at start; a lazy `GtfsShape.points` view over the typed columns would cut decode time and memory further (357 k objects are allocated today, as the parser does).

### DIFF
--- a/packages/trufi_core_planner/lib/src/index/gtfs_route_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_route_index.dart
@@ -209,6 +209,11 @@ class PatternConnections extends ListBase<PatternConnection> {
 }
 
 /// Index for fast route lookups.
+///
+/// Persisted by the planner index snapshot (#993): whenever a change here
+/// makes this build different patterns, connections or line keys from the
+/// same GTFS, bump `PlannerIndexCodec.formatVersion`, or users keep the
+/// previous release's index until the bundled feed changes.
 class GtfsRouteIndex {
   /// Default radius within which two distinct stops count as one transfer
   /// point. Matches MOTIS' `link_stop_distance` (100 m straight-line);
--- a/packages/trufi_core_planner/lib/src/index/gtfs_route_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_route_index.dart
@@ -268,6 +268,81 @@ class GtfsRouteIndex {
     );
   }
 
+  /// Restores an index from a snapshot of one built by [GtfsRouteIndex.new]
+  /// over the same [data] with the same knobs: the [patterns] in id order
+  /// and the CSR connection columns exactly as [connectionStarts],
+  /// [connectionOtherPattern], [connectionMyStopIdx], [connectionOtherStopIdx]
+  /// and [connectionWalk] returned them. The cheap per-stop and per-route
+  /// maps are re-derived from the patterns: iterating them in id order
+  /// inserts every (stop, pattern) and (stop, route) pair in the same order
+  /// the build did — a pattern id is the order of first appearance of its
+  /// trip, so the first trip of a route that visits a stop is also its first
+  /// pattern there. Used by the planner index snapshot; the codec validates
+  /// the arrays before calling this.
+  GtfsRouteIndex.restore({
+    required GtfsData data,
+    required List<RoutePattern> patterns,
+    required Int32List connectionStarts,
+    required Int32List connectionOtherPattern,
+    required Int32List connectionMyStopIdx,
+    required Int32List connectionOtherStopIdx,
+    required Float32List connectionWalk,
+    required this.transferRadiusMeters,
+    required this.sameNameRouteLimit,
+  }) : _data = data {
+    if (connectionStarts.length != patterns.length + 1) {
+      throw ArgumentError(
+        'connectionStarts must have patterns.length + 1 entries',
+      );
+    }
+    final total = connectionStarts.isEmpty ? 0 : connectionStarts.last;
+    if (connectionOtherPattern.length != total ||
+        connectionMyStopIdx.length != total ||
+        connectionOtherStopIdx.length != total ||
+        connectionWalk.length != total) {
+      throw ArgumentError('connection columns must have $total entries');
+    }
+    for (var i = 0; i < patterns.length; i++) {
+      if (patterns[i].id != i) {
+        throw ArgumentError('pattern at position $i has id ${patterns[i].id}');
+      }
+    }
+    _patterns = patterns;
+    _routePatterns = {};
+    _stopToPatternIds = {};
+    _stopToRoutes = {};
+    for (final pattern in patterns) {
+      _routePatterns.putIfAbsent(pattern.routeId, () => []).add(pattern);
+      for (final stopId in pattern.stopIds) {
+        _stopToRoutes.putIfAbsent(stopId, () => {}).add(pattern.routeId);
+        _stopToPatternIds.putIfAbsent(stopId, () => {}).add(pattern.id);
+      }
+    }
+    _buildLineKeys();
+    _connStart = connectionStarts;
+    _connOther = connectionOtherPattern;
+    _connMyIdx = connectionMyStopIdx;
+    _connOtherIdx = connectionOtherStopIdx;
+    _connWalk = connectionWalk;
+  }
+
+  /// CSR offsets of the connection table: pattern `p` owns entries
+  /// `[connectionStarts[p], connectionStarts[p + 1])` of the four columns
+  /// below. Exposed for the snapshot codec; treat all five as read-only.
+  Int32List get connectionStarts => _connStart;
+
+  /// Column: id of the pattern boarded (see [PatternConnection.otherPatternId]).
+  Int32List get connectionOtherPattern => _connOther;
+
+  /// Column: alight position in the source pattern.
+  Int32List get connectionMyStopIdx => _connMyIdx;
+
+  /// Column: boarding position in the other pattern.
+  Int32List get connectionOtherStopIdx => _connOtherIdx;
+
+  /// Column: straight-line walk between alight and boarding stop, meters.
+  Float32List get connectionWalk => _connWalk;
+
   void _buildIndices() {
     _patterns = [];
     _stopToPatternIds = {};
--- a/packages/trufi_core_planner/lib/src/index/gtfs_schedule_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_schedule_index.dart
@@ -60,6 +60,32 @@ class GtfsScheduleIndex {
     _buildIndices();
   }
 
+  /// Restores an index from the already grouped and sorted
+  /// [stopTimesByStop] of an index built over the same data (see
+  /// [stopTimesByStop]); the per-route trip lists are re-derived from
+  /// [trips], whose order is preserved by the snapshot. Used by the planner
+  /// index snapshot.
+  GtfsScheduleIndex.restore({
+    required Map<String, GtfsTrip> trips,
+    required List<GtfsStopTime> stopTimes,
+    required Map<String, GtfsCalendar> calendars,
+    required List<GtfsCalendarDate> calendarDates,
+    required List<GtfsFrequency> frequencies,
+    required Map<String, List<GtfsStopTime>> stopTimesByStop,
+  }) : _trips = trips,
+       _stopTimes = stopTimes,
+       _calendars = calendars,
+       _calendarDates = calendarDates,
+       _frequencies = frequencies {
+    _stopTimesByStop = stopTimesByStop;
+    _indexTripsByRoute();
+  }
+
+  /// Stop times grouped by stop id, each list sorted by departure time —
+  /// the index's own working table, exposed for the snapshot codec. Treat
+  /// as read-only.
+  Map<String, List<GtfsStopTime>> get stopTimesByStop => _stopTimesByStop;
+
   void _buildIndices() {
     // Index stop times by stop ID
     _stopTimesByStop = {};
@@ -67,11 +93,7 @@ class GtfsScheduleIndex {
       _stopTimesByStop.putIfAbsent(st.stopId, () => []).add(st);
     }
 
-    // Index trips by route ID
-    _tripsByRoute = {};
-    for (final trip in _trips.values) {
-      _tripsByRoute.putIfAbsent(trip.routeId, () => []).add(trip);
-    }
+    _indexTripsByRoute();
 
     // Sort stop times by departure time
     for (final stopTimes in _stopTimesByStop.values) {
@@ -80,6 +102,13 @@ class GtfsScheduleIndex {
         final timeB = b.departureTime ?? b.arrivalTime ?? Duration.zero;
         return timeA.compareTo(timeB);
       });
+    }
+  }
+
+  void _indexTripsByRoute() {
+    _tripsByRoute = {};
+    for (final trip in _trips.values) {
+      _tripsByRoute.putIfAbsent(trip.routeId, () => []).add(trip);
     }
   }
 
--- a/packages/trufi_core_planner/lib/src/index/gtfs_schedule_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_schedule_index.dart
@@ -34,6 +34,11 @@ class RouteFrequencyInfo {
 }
 
 /// Index for schedule-based queries (departures, frequencies).
+///
+/// Persisted by the planner index snapshot (#993): whenever a change here
+/// makes this build different groups or order from the same GTFS, bump
+/// `PlannerIndexCodec.formatVersion`, or users keep the previous release's
+/// index until the bundled feed changes.
 class GtfsScheduleIndex {
   final Map<String, GtfsTrip> _trips;
   final List<GtfsStopTime> _stopTimes;
--- a/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
@@ -13,6 +13,11 @@ class NearbyStop {
 }
 
 /// Spatial index for fast nearest-stop queries using a KD-Tree.
+///
+/// Persisted by the planner index snapshot (#993): whenever a change here
+/// makes this build a different tree from the same GTFS, bump
+/// `PlannerIndexCodec.formatVersion`, or users keep the previous release's
+/// index until the bundled feed changes.
 class GtfsSpatialIndex {
   final Map<String, GtfsStop> _stops;
   late final _KdTree _tree;
--- a/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
+++ b/packages/trufi_core_planner/lib/src/index/gtfs_spatial_index.dart
@@ -21,6 +21,19 @@ class GtfsSpatialIndex {
     _tree = _KdTree(_stops.values.toList());
   }
 
+  /// Restores an index from the pre-order traversal of a tree built by
+  /// [GtfsSpatialIndex.new] over the same stops (see [preorder]). The tree
+  /// shape is a function of the stop count alone (median split), so the
+  /// result is the very same tree — same answers, same result order —
+  /// without sorting anything. Used by the planner index snapshot.
+  GtfsSpatialIndex.restore(this._stops, List<GtfsStop> preorder) {
+    _tree = _KdTree.fromPreorder(preorder);
+  }
+
+  /// The tree's stops in pre-order (node, left subtree, right subtree);
+  /// feeds [GtfsSpatialIndex.restore].
+  List<GtfsStop> get preorder => _tree.preorder();
+
   /// Find the nearest stops to a location.
   List<NearbyStop> findNearestStops(
     LatLng location, {
@@ -79,6 +92,39 @@ class _KdTree {
     if (stops.isNotEmpty) {
       _root = _buildTree(stops, 0);
     }
+  }
+
+  /// Rebuilds the tree from its pre-order listing. [_buildTree] puts the
+  /// median (`length ~/ 2`) at the node with the `mid` smaller stops on the
+  /// left, so the left subtree of a node covering `count` stops always has
+  /// `count ~/ 2` of them.
+  _KdTree.fromPreorder(List<GtfsStop> preorder) {
+    if (preorder.isNotEmpty) {
+      _root = _fromPreorder(preorder, 0, preorder.length);
+    }
+  }
+
+  _KdNode? _fromPreorder(List<GtfsStop> stops, int start, int count) {
+    if (count == 0) return null;
+    final leftCount = count ~/ 2;
+    return _KdNode(
+      stop: stops[start],
+      left: _fromPreorder(stops, start + 1, leftCount),
+      right: _fromPreorder(stops, start + 1 + leftCount, count - leftCount - 1),
+    );
+  }
+
+  List<GtfsStop> preorder() {
+    final out = <GtfsStop>[];
+    void visit(_KdNode? node) {
+      if (node == null) return;
+      out.add(node.stop);
+      visit(node.left);
+      visit(node.right);
+    }
+
+    visit(_root);
+    return out;
   }
 
   _KdNode? _buildTree(List<GtfsStop> stops, int depth) {
--- a/packages/trufi_core_planner/lib/src/parser/csv_parser.dart
+++ b/packages/trufi_core_planner/lib/src/parser/csv_parser.dart
@@ -1,4 +1,9 @@
 /// Simple CSV parser for GTFS files.
+///
+/// Persisted by the planner index snapshot (#993): whenever a change here
+/// makes this produce different rows from the same GTFS, bump
+/// `PlannerIndexCodec.formatVersion`, or users keep the previous release's
+/// parse until the bundled feed changes.
 class CsvParser {
   /// Parse CSV content into a list of maps.
   /// Each map represents a row with column names as keys.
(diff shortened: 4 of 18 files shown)


SamuelRioTz:
## Review — measured on the real Sana'a app, verified both ways

**Verdict: works as described; fresh-eyes review done and its findings applied — recommend merge.** The offline planner used to parse the GTFS and build every index on each cold start; this branch snapshots the whole in-memory state (parsed data + the three indices) in the app's cache directory and restores it on the next start, rebuilding only when the GTFS content, the index knobs or the snapshot format version differ — or when the file is not intact.

- **Design checked against the measurements, not the issue text.** On the feed trufi-app ships, parsing is 68 % of the cold start (`shapes.txt` alone 45 %) and the #992 transfer table 25 %, so persisting only the indices — the issue's literal proposal — would have left two thirds of the cost in place. The snapshot holds the parsed `GtfsData` too, losslessly (`Float64` coordinates, `Int32` seconds, dates as µs + UTC flag, maps in iteration order).
- **Code.** `PlannerIndexCodec` is pure Dart over `dart:typed_data`: 8-byte-aligned tagged sections, a string pool written as the last section (its offset in the header) so the encoder streams into one buffer, integer columns at their narrowest width, the CSR connection table verbatim, the KD-tree as its pre-order (same shape ⇒ same tree, no sort), the sorted schedule groups. Header = magic, `formatVersion`, byte-order marker, string pool offset, knobs, GTFS fingerprint (two 32-bit FNV-1a streams over words, JS-safe as in #980), payload length + checksum, trailer. `decode` refuses anything it will not vouch for — every header bit has its own check, record counts are bounded against the remaining bytes before anything is allocated — and never throws for malformed input. The three indices gain `restore` constructors; the build path is untouched. In routing, `path_provider` + `dart:io` sit behind a conditional import (web keeps building; `flutter build web --release` of `apps/example` passes), the path is resolved on the root isolate, the worker reads/decodes or builds/encodes, and the write (unique temp file + rename, temp removed on failure) happens after the planner is ready. `TrufiPlannerConfig.local(persistIndex:)` opts out.
- **Cache directory on purpose** (not application support): Android Auto Backup caps an app at 25 MB and excludes `getCacheDir()`; a 38 MB file under `getFilesDir()` would make the whole app backup fail. iOS keeps `Library/Caches` out of iCloud and may purge it — the snapshot is rebuilt.
- **Desktop harness, both feeds:** Cochabamba build 1.63–1.70 s → decode **0.17–0.32 s** (snapshot 38.25 MB, 7.8× the zip: 26 MB are the 2.6 M #992 connections, 7 MB the 357 k shape points), Sana'a 0.11–0.19 s → **8–31 ms** (1.12 MB). Re-encoding the restored bundle is byte-identical; a canonical dump of every field (3.1 M lines on Cochabamba) shows 0 differences between the built and the restored bundle; 300 random pairs × 2 walk radii per feed plan identical itineraries. First-start peak on the desktop: build alone 620–628 MB maxRss, build + encode 687–697 MB (was 763–775 MB before the single-buffer encoder), decode in a fresh process 444 MB.
- **Tests:** planner **85/85** (56 before), routing unit **140/140** (126 before). The codec suite covers every field of the round trip, the order of per-stop lookups, all four connection columns, 72 truncation points, 300 random bit flips, every bit of the header, forged counts and offsets with a valid checksum, other version/knobs/fingerprint, and pins a digest of **every field** of the fixture bundle — parsed data included — so a parser or index change without a `formatVersion` bump fails the build (the #973 failure mode; the rule is documented on `formatVersion`, the parser and the three indices). The routing suite exercises real files in a temp cache dir through the real isolate: build → write → load → identical plans, changed asset → rebuild and the new data used, corrupt / truncated / garbage / other-version → silent rebuild, opt-out, missing `path_provider`, the planner ready before the write starts (write held at a test seam), the write atomic (temp file next to the target, renamed — not copied — over the previous snapshot, nothing left behind, temp removed on failure).
- **Fresh-eyes review** (read-only pass with its own harness on both feeds and 21 mutations of the branch): the codec restored 0 differences everywhere and never threw on 100 000+ corrupted inputs; its three findings — the guard covered only the integer skeleton, the encoder's second writer cost ~150 MB of transient peak, and no test pinned the atomic write or the ready-before-write ordering — are fixed in the two follow-up commits and re-verified by the same mutations (all four that used to survive are now caught).

### Evidence — trufi-sanaa app on an Android 16 emulator, **profile (AOT)** build, `dependency_overrides` to `main` (`3ac4838d`) and to this branch

Times from logcat: *planner ready* = `Preloaded in` (from the start of `_preloadLocal`), the isolate's own split in parentheses.

| Scenario | Planner ready | Inside the isolate |
|---|---|---|
| `main`, cold start 1 / 2 | **476 / 642 ms** | parse + build, every start; no `cache/trufi_planner/` |
| branch, first start (no snapshot) | 610 ms | `built in 584ms` (parse 311, spatial 33, index 220, schedule 1; encode 16) + async write 1.1 MB in 126 ms |
| branch, cold start 2 / 3 | **107 / 81 ms** | `loaded from cache in 29ms` (decode 28) / `43ms` (decode 39) |
| GTFS without any line "14" (29 route_ids) installed over it, cache kept | 523 ms | `cache rejected: GTFS fingerprint changed` → built in 496 ms (parse 319, spatial 79, index 20; encode 68); snapshot 1.0 MB (967 012 B, was 1 121 572), snapshot rewritten; the university trip (7 → 14 before) finds no route on that feed |
| next cold start on the new feed | 72 ms | loaded from cache in 52 ms (decode 50) |
| snapshot corrupted (4 KB overwritten with 0xFF) | 581 ms | `cache rejected: checksum mismatch` → `built in 567ms`, rewritten; next start **42 ms** from cache |
| snapshot truncated to 300 KB | 463 ms | `cache rejected: truncated snapshot (no trailer)` → `built in 428ms`, rewritten |

Debug (JIT) builds keep the ratio but run slower: `main` 1 594 / 1 259 ms; branch first start 1 516 ms, then **598 / 631 ms** from the cache. 0 `Unhandled Exception|FATAL|Null check` across all runs.

```
# main (3ac4838d)
I/flutter: TrufiPlannerDataSource: Preloaded in 476ms
I/flutter: TrufiPlannerDataSource: Preloaded in 642ms
# branch, first start
I/flutter: TrufiPlannerDataSource: Preloaded in 610ms — planner index built in 584ms (parse 311ms, spatial 33ms, index 220ms, schedule 1ms; snapshot 1.1 MB encoded in 16ms; fingerprint e375b10d3d0bb037)
I/flutter: TrufiPlannerDataSource: planner index snapshot written (1.1 MB in 126ms) to /data/user/0/app.trufi.sanaa/cache/trufi_planner/sanaa.gtfs.zip-ec63c8f45c24eb6b.idx
# branch, next cold starts
I/flutter: TrufiPlannerDataSource: Preloaded in 107ms — planner index loaded from cache in 29ms (read 0ms, decode 28ms; snapshot 1.1 MB, fingerprint e375b10d3d0bb037)
I/flutter: TrufiPlannerDataSource: Preloaded in 81ms — planner index loaded from cache in 43ms (read 2ms, decode 39ms; snapshot 1.1 MB, fingerprint e375b10d3d0bb037)
# corrupt / truncated snapshot
I/flutter: TrufiPlannerDataSource: Preloaded in 581ms — planner index built in 567ms (…; cache rejected: checksum mismatch; …)
I/flutter: TrufiPlannerDataSource: Preloaded in 463ms — planner index built in 428ms (…; cache rejected: truncated snapshot (no trailer); …)
# feed without any line "14" installed over it (cache from the original feed still on disk)
I/flutter: TrufiPlannerDataSource: Preloaded in 523ms — planner index built in 496ms (parse 319ms, spatial 79ms, index 20ms, schedule 0ms; snapshot 1.0 MB encoded in 68ms; cache rejected: GTFS fingerprint changed; fingerprint 14c9269edea318d2)
I/flutter: TrufiPlannerDataSource: planner index snapshot written (1.0 MB in 33ms) to /data/user/0/app.trufi.sanaa/cache/trufi_planner/sanaa.gtfs.zip-ec63c8f45c24eb6b.idx
I/flutter: TrufiPlannerDataSource: Preloaded in 72ms — planner index loaded from cache in 52ms (read 0ms, decode 50ms; snapshot 1.0 MB, fingerprint 14c9269edea318d2)
```

| University trip (trufi-sanaa#2 B-fwd) planned **from the cached snapshot**: 7 → 14, 57 min — same as `main` | Same trip after installing the feed **without any line "14"** over it: fingerprint rebuild, **no route on that feed** |
|---|---|
|  |  |

Cochabamba could not be run in the emulator (trufi-app's toolchain, as in #992); its numbers come from the harness on the feed trufi-app ships.

### Notes / follow-ups (not blockers)

1. First start after install or update pays build + encode (Sana'a +16 ms AOT; Cochabamba +0.24–0.37 s and ~65–75 MB of peak RSS on the desktop) before the planner is ready; the write itself is asynchronous (pinned by a test).
2. Snapshot size: 38 MB for Cochabamba in the cache dir. `gzip -1` would make it 10 MB (0.22 s to compress once, 0.03 s to inflate on the desktop) — left out of v1 to keep the warm path a plain read; numbers in the PR.
3. One snapshot file per asset name: an app that renames its GTFS asset leaves the old `.idx` in the cache until the system purges it. Removing "everything but mine" would break an app with two local assets, so this needs a small policy — follow-up.
4. Calendar dates are persisted as the local-time instants the parser creates; nothing reads them yet. When the schedule starts using them, persist year-month-day or add the timezone to the key.
5. A lazy `GtfsShape.points` view over the typed columns would cut decode time and memory further (357 k point objects are allocated today, as the parser does).
6. `GtfsScheduleIndex` is persisted although nothing reads it yet (~0.5 MB on Cochabamba).

PR #991, opened 2026-09-09
Copy place and stop names via long-press; copy button on the POI card (trufi-sanaa#9)

SamuelRioTz:
Implements trufi-association/trufi-sanaa#9. In Sana'a, passing a place name on (mostly Arabic) meant a screenshot plus OCR: every name in the app is a plain `Text` inside an `InkWell.onTap` that *selects* the place, so nothing could be copied. Map labels themselves are MapLibre tiles and cannot be selected; the existing long-press-on-map → origin/destination flow puts the name into a field, and that field now copies.

## What changes

**Long-press on any place or stop name copies it** — the name only, never the "name, address" a field renders — with a light haptic tick and a short floating "Copied" SnackBar. Tap behaviour is untouched on every surface.

| Surface | Where |
|---|---|
| Search results rows and "Your places" rows | `trufi_core_search_locations/…/location_search_screen.dart` |
| Home origin/destination fields, compact and wide variants | `trufi_core_search_locations/…/search_location_bar.dart` |
| POI card — both the bottom panel of the narrow layout (`POIDetailPanel`) and the compact card of the wide side panel (tablets, phones in landscape, desktop web): a **copy button** next to ✕, plus long-press on the name and on the address (and the hours / phone rows the narrow panel has) | `trufi_core_poi_layers/…/poi_detail_panel.dart`, `trufi_core_home_screen/…/poi_compact_card.dart` |
| Saved place tile | `trufi_core_saved_places/…/saved_place_tile.dart` |
| Itinerary timeline: origin, transfers, destination, expanded intermediate stops | `trufi_core_home_screen/…/itinerary_detail_screen.dart` |
| Route detail: stop rows and the origin/destination header (bottom sheet and side panel) | `trufi_core_transport_list/…/transport_detail_screen.dart` |

A corner opened from its street's list shows the trimmed "& Avenida Aroma" label but copies the full "Avenida Ayacucho & Avenida Aroma".

The wide layout's POI card used to be inlined in `home_screen.dart` (`_buildPOIDetailPanel(isWideScreen: true)`); it is now `POICompactCard`, a public widget of `trufi_core_home_screen` — the same pattern as `LiveVehicleInfoPanel` — so it can be widget-tested without pumping the whole home screen. Its layout is unchanged; the copy button shares the ✕'s constraints so the two stay aligned.

[...]

PR #996, opened 2026-09-10
Home: draw and fit a plan restored from the previous session (#995)

### LINKED ISSUE #995: Home: a plan restored from the previous session is not drawn on the map (no polyline, no markers, no camera fit)
## Problem

When the app is cold-started with a plan from the previous session, the home screen restores the itinerary panel ("1 route found", duration, legs, distance) but the **map shows nothing of it**: no route polyline, no origin/destination markers, no leg badges, and the camera stays at the default city center instead of fitting the route. The panel and the map disagree until the user triggers a new search.

Reproduced on the real Sana'a app (Android 16 emulator, profile build) on `main` at `3ac4838d` and again at `61eac621`: plan the university trip via the `trufiapp://route?…` deep link (route drawn correctly), force-stop, cold start with data kept → panel restored, map empty, at 3 s and still at 15 s. Same result with and without the persisted planner index (#994), so this is not a timing issue with the planner — it is the restore path itself.

| after the deep link | cold start, plan restored |
|---|---|
|  | (panel "1 route found · 57 min · 7 → 14" visible, map at default center, no polyline, no markers) |

## Where to look

- `HomeScreenRepository` persists the last plan and the home cubit restores it on startup (see 5.14.1: `Leg.serviceHours` lost on reload — the restore path is exercised on every relaunch).
- The map layers for the selected itinerary and the origin/destination markers are driven by the plan state; whatever draws them after a fresh `fetchPlan` is not run (or runs before the map engine/style is ready and is not replayed) when the plan comes from the repository. Related precedents: #946 (camera replay after style load), #902/#903 (viewport reset on theme change / POI set).

## Expected

On cold start with a restored plan: the selected itinerary is drawn (polyline + leg badges), origin/destination markers are shown, and the camera fits the itinerary bounds — exactly as after a fresh search. If the plan is stale (e.g. departure time in the past, or the GTFS changed), either re-plan or clear both the panel and the map consistently; never show a panel whose route is not on the map.

## Acceptance criteria

- Emulator evidence before/after on the Sana'a app: cold start with a restored plan draws the route and fits the camera; a new search still works; clearing the plan clears both.
- Covers the map-not-ready case: the plan restore may complete before the map style loads — the draw must be replayed once the map is ready (widget test with a fake map engine, as in `trufi_core_transport_list`'s tests).
- No change for the fresh-search path.

SamuelRioTz:
Closes #995. On a cold start with a plan from the previous session the home screen restored the itinerary panel but the map stayed at the default city view. The route **was** on the map — `TrufiMap` synchronised the polyline and marker layers as soon as its style loaded — but the camera never fitted it, and the university trip of trufi-sanaa#2 lies entirely outside Sana'a's default viewport (centre 15.347/44.205, zoom 14), so the screen looked empty. Two swipes south on the "empty" map showed the "14" line and its badge.

The issue title says "no polyline, no markers"; the instrumented run shows both layers were synced to the native map (`route-lines-layer` 4 lines, `route-markers-layer` 8 markers, including the origin/destination markers) — the only missing piece was the camera fit. The issue title and body should be read as "the restored route is drawn outside the default viewport and the camera does not fit it".

## Root cause (timeline from an instrumented profile build of the Sana'a app, Android 16)

```
09.129  first frame: _viewportReady = true, no pending fit points
09.132  RoutePlannerCubit.initialize → plan restored (4 legs, 167 points) → listener → route layers built
09.132  _fitCameraToPoints: _currentCamera == null → return           ← fit dropped
09.581  native camera idle for the initial position (before the style)
09.597  first onCameraChanged (default centre), pendingFit = 167       ← nothing replays the fit
09.754  onStyleLoaded → layers synced (route-lines-layer 4 lines, route-markers-layer 8 markers)
```

- `_fitCameraToPoints` returned early when the map had not reported a camera yet (`_currentCamera` is only set by `onCameraChanged` / `_moveCameraTracked`). On a cold start the restore lands ~3 ms after the first frame and ~465 ms before the map's first camera report.
- The only replay of `_pendingFitPoints` runs once, in the first `LayoutBuilder` build — 3 ms *before* the restore, with nothing to replay — and never again. A fresh search never hits this because the user searches after the map has reported its camera.
- The #946 replay in `TrufiMap` could not help: no controlled camera was ever produced.
- The JSON round trip is not involved: the local planner stores `encodedPoints` on every leg and the restored legs decode to their full geometry (checked in the log and pinned by the tests).

## Fix

- **`home_screen.dart` — `_fitCameraToPoints`**: use `_initialCamera` as the base camera while the map has not reported one. `FitCameraUtil.cameraForPoints` takes only the bearing from the base camera (`TrufiCameraPosition` has no tilt, and the base zoom cancels out of the fit maths: target and zoom come from the points and the viewport, known since the first frame), so the fit is computed as soon as the plan is there and reaches `TrufiMap` as a controlled camera — reusing the #946 mechanism, which holds it until the style has loaded. No new replay path, no waiting on events.
- **`trufi_map.dart` — `_handleCameraIdle`**: ignore camera-idle events while a controlled camera is pending for the style load. Android fires `camera#onIdle` for the initial position before `onStyleLoaded` (registered in `onMapReady`, before `setStyle`); without this guard that idle overwrote `_currentCamera` with the default centre and the replay in `onStyleLoadedCallback` became an "already there" no-op — the fit would have been lost a second way. This closes a gap in #946 for any controlled camera set before the style loads; a pan made during the (sub-second) style load is overridden by the requested camera, which is what a controlled camera means.
- **`home_screen.dart` — `_moveCameraTracked`** (found in review): drop `_cameraTarget` when the camera is moved programmatically (my-location). The map suppresses its callback for programmatic moves, so `_onCameraChanged` never cleared the driven camera and the next rebuild (a GPS tick, the keyboard, a rotation) handed the route fit back to the map, snapping the view away from the user's location. Pre-existing on the fresh-search path (search → my-location without panning first); with the restored fit it would have hit the first my-location tap of every cold start, and it also defeated #950 (an engine switch after my-location would have reopened on the route).

The fresh-search path is unchanged: with a reported camera the code takes exactly the branch it took before. Clearing the plan still clears layers and camera together (`_updateRouteOnMap(null)` / `_updateFitCameraPoints`).

Stale plans are drawn as-is and fitted, as before this PR nothing re-planned or dropped a restored plan; a "from the previous session / re-plan" hint would be a product follow-up, not part of this bug.

## Tests (Flutter 3.44.1, the CI version)

`trufi_core_home_screen` **39/39** (31 before this PR), `trufi_core_maps` **56/56** (50 before), `flutter analyze` clean in both packages apart from a pre-existing info in an untouched file.

`test/restored_plan_map_test.dart` (+8, `trufi_core_home_screen`) drives the **real `HomeScreen`** over a fake map engine that records the `layers` / `camera` / `initialCamera` it is handed and can play the map's camera report; the repository stores JSON like the real one:

- restored **after the first frame and before any camera report** (the device timeline): 4 lines, start/end/transit-label/boarding/alighting markers, camera fitted inside the trip's bounding box, zoomed out from 14, bearing kept at 0;
- restored **before the screen exists** (state already set when the `BlocConsumer` subscribes — the `needsRouteRender` fallback);
- wide layout (side panel) fits too;
- the map reporting the fitted camera afterwards: the screen releases the controlled camera (a later rebuild hands none) and the map keeps the fitted view;
- `clearPlan()` removes lines and markers, drops the camera target and leaves the origin/destination previews;
- **my-location after the fit**: a real tap on the my-location button over a mocked geolocator (tracking auto-started), then a rebuild — the map stays on the user and the screen hands no camera back;
- panning away shows the recentre button; fitting another itinerary hides it again;
- fresh search after the map reported its camera draws and fits (unchanged).

The same file on `main`: ****7 of the 8 fail** (every restore-based scenario — `camera` is null, so the my-location and recentre steps never get a fit either); the fresh-search guard passes**. Each guard has a mutation check: dropping the `_cameraTarget = null` line of `_moveCameraTracked` fails exactly the my-location test (`+38 −1`); reverting `_handleCameraIdle` to `main` leaves this suite green and fails exactly 3 of the 6 tests below (`+53 −3` in `trufi_core_maps`).

`test/trufi_map_idle_guard_test.dart` (+6, `trufi_core_maps`) drives the **real `TrufiMap`** over a fake `MapLibrePlatform` (the plugin's `MapLibrePlatform.createInstance` seam, no platform view) and plays the native event order by hand: controlled camera → initial `camera#onIdle` → `map#onStyleLoaded`. On `main` 3 of the 6 fail (the initial idle is reported and the replay becomes a no-op). The other three pin what must not change: the #946 order (idle → camera → style), an inert guard without a controlled camera, and a controlled camera set after the style (immediate move, own idle swallowed, next gesture reported).

## Emulator (Sana'a app, Android 16 arm64, profile builds, `dependency_overrides` to `main` `61eac621` and to this branch)

Sequence in both builds: deep link of the university trip (route drawn) → force-stop → cold start with data kept → screenshots at 3 s and 15 s.

| | after the deep link (fresh path) | cold start + 3 s | + 15 s |
|---|---|---|---|
| `main` | route drawn | panel "1 route found · 57 min" restored, map at the default centre, no route in view | identical bytes |
| this branch | route drawn | panel restored, **route drawn and camera fitted** (7 → 14, origin and destination markers) | identical bytes |

On the branch, additionally: rotate to landscape (wide layout, side panel) → route still fitted in the map area; back to portrait → kept; tap the panel's ✕ → route and markers gone together with the panel; cold start after that → nothing restored, default view. 0 crash lines in logcat across all runs.

**My-location snap-back (the `_moveCameraTracked` fix), same emulator, restored plan on both builds**: cold start with the restored plan → a GPS fix delivered → tap the my-location button → 2 s later a screenshot → one more GPS fix (a `HomeScreen` rebuild) → 5 s later a screenshot. On `a7ba58be` the map is on the user (zoom 16) at 2 s and **back on the fitted route** at 5 s; with the fix it is on the user at 2 s and **still on the user** at 5 s (the dot moved with the new fix, so the tick was delivered). 0 crash lines.

| `a7ba58be` — 2 s after my-location | `a7ba58be` — 5 s later, after a GPS tick: snapped back to the route | this branch — 2 s after my-location | this branch — 5 s later, after a GPS tick: stays on the user |
|---|---|---|---|
|  |  |  |  |

| `main` — cold start + 15 s (panel restored, map at the default centre) | `main` — same screen after panning south: the route was there, off-screen | this branch — cold start + 3 s: restored route drawn and fitted |
|---|---|---|
|  |  |  |

The full set (deep link on both builds, 15 s, rotation, clearing, cold start after clearing) is in the review comment.

## Notes for the reviewer

- `home_screen.dart` changes 4 lines of logic plus two doc comments (`_fitCameraToPoints`, `_moveCameraTracked`); #991 (open) touches the same file only around the POI panel (`_buildPOIDetailPanel`, imports) — no overlapping hunks, either can merge first.
- `_currentCamera` in `HomeScreen` stays null until the map's first report, as before; `_buildMap` already starts a rebuilt map (engine or theme switch) from `camera: _cameraTarget`, which after this PR is the fitted camera, so the view survives an engine switch; the rotation check above exercises a rebuild of the same map, not an engine switch.
- `dart format` on `trufi_map.dart` would reflow lines `main` never formatted; only the touched lines are formatted, as in previous PRs.

### DIFF
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -1026,15 +1026,12 @@ class _HomeScreenState extends State<HomeScreen>
 
   /// Compute and apply camera position to fit the given points.
   ///
-  /// The fit only takes bearing and tilt from the base camera — target and
-  /// zoom come from the points — so before the map has reported a camera the
-  /// initial one is used instead of dropping the fit. That is the cold-start
-  /// case with a plan restored from the previous session: the plan is back a
-  /// few milliseconds after the first frame, hundreds of milliseconds before
-  /// the map's first camera report, and nothing replayed the fit afterwards,
-  /// so the restored route was drawn outside the default viewport (#995).
-  /// The controlled camera set here is held by [TrufiMap] until its style has
-  /// loaded (#946), so nothing needs to wait for the map.
+  /// The fit only takes the bearing from the base camera — target and zoom
+  /// come from the points and the viewport — so before the map has reported a
+  /// camera the initial one is used instead of dropping the fit: a plan
+  /// restored on a cold start lands before the map's first camera report and
+  /// nothing replayed the fit afterwards (#995). The controlled camera set
+  /// here is held by [TrufiMap] until its style has loaded (#946).
   void _fitCameraToPoints(List<LatLng> points) {
     final baseCamera = _currentCamera ?? _initialCamera;
     if (_fitCameraUtil == null || baseCamera == null) return;
@@ -1081,9 +1078,14 @@ class _HomeScreenState extends State<HomeScreen>
   /// Programmatic moves suppress the map's onCameraChanged callback, so
   /// without this the recorded viewport goes stale and an engine switch
   /// right after (e.g. my-location → change map style) snaps back to the
-  /// previous view (#902).
+  /// previous view (#902). For the same reason the controlled camera is
+  /// dropped here: [_onCameraChanged] never runs for this move, so it would
+  /// stay set to the route fit and the next rebuild (a GPS tick, the
+  /// keyboard, a rotation) would hand it back to [TrufiMap], snapping the
+  /// view away from the location the user asked for (#995).
   void _moveCameraTracked(TrufiCameraPosition position) {
     _currentCamera = position;
+    _cameraTarget = null;
     _mapController?.moveCamera(position);
   }
 
--- a/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
+++ b/packages/screens/trufi_core_home_screen/lib/src/widgets/home_screen.dart
@@ -1025,9 +1025,20 @@ class _HomeScreenState extends State<HomeScreen>
   }
 
   /// Compute and apply camera position to fit the given points.
+  ///
+  /// The fit only takes bearing and tilt from the base camera — target and
+  /// zoom come from the points — so before the map has reported a camera the
+  /// initial one is used instead of dropping the fit. That is the cold-start
+  /// case with a plan restored from the previous session: the plan is back a
+  /// few milliseconds after the first frame, hundreds of milliseconds before
+  /// the map's first camera report, and nothing replayed the fit afterwards,
+  /// so the restored route was drawn outside the default viewport (#995).
+  /// The controlled camera set here is held by [TrufiMap] until its style has
+  /// loaded (#946), so nothing needs to wait for the map.
   void _fitCameraToPoints(List<LatLng> points) {
-    if (_fitCameraUtil == null || _currentCamera == null) return;
-    final newCam = _fitCameraUtil!.cameraForPoints(points, _currentCamera!);
+    final baseCamera = _currentCamera ?? _initialCamera;
+    if (_fitCameraUtil == null || baseCamera == null) return;
+    final newCam = _fitCameraUtil!.cameraForPoints(points, baseCamera);
     if (newCam != null) {
       setState(() {
         _cameraTarget = newCam;
--- a/packages/screens/trufi_core_home_screen/test/restored_plan_map_test.dart
+++ b/packages/screens/trufi_core_home_screen/test/restored_plan_map_test.dart
@@ -1,0 +1,532 @@
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:latlong2/latlong.dart';
+import 'package:provider/provider.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+import 'package:trufi_core_home_screen/trufi_core_home_screen.dart';
+import 'package:trufi_core_interfaces/trufi_core_interfaces.dart'
+    show TrufiLocation;
+import 'package:trufi_core_maps/trufi_core_maps.dart';
+import 'package:trufi_core_routing/trufi_core_routing.dart' as routing;
+import 'package:trufi_core_search_locations/trufi_core_search_locations.dart'
+    show SearchLocationsLocalizations;
+
+/// #995: a plan restored from the previous session has to reach the map the
+/// way a fresh one does — route layers AND a fitted camera — although the
+/// restore completes before the map has reported any camera. Drives the real
+/// [HomeScreen] over a fake map engine that records what it is handed.
+
+// Sana'a: the app's default viewport vs the university trip (trufi-sanaa#2),
+// which lies entirely outside that viewport.
+const _defaultCenter = LatLng(15.3470, 44.2050);
+const _defaultZoom = 14.0;
+const _origin = LatLng(15.2952334, 44.2623529);
+const _boarding = LatLng(15.2990, 44.2580);
+const _transfer = LatLng(15.3150, 44.2250);
+const _alighting = LatLng(15.3500, 44.1830);
+const _destination = LatLng(15.3536032, 44.1786948);
+
+routing.Place _place(String id, LatLng p) =>
+    routing.Place(name: id, lat: p.latitude, lon: p.longitude, stopId: id);
+
+routing.Leg _leg(
+  List<LatLng> points, {
+  String? route,
+  required int startMinute,
+  required int minutes,
+}) {
+  final start = DateTime(2026, 9, 10, 12, startMinute);
+  return routing.Leg(
+    mode: route == null ? 'WALK' : 'BUS',
+    startTime: start,
+    endTime: start.add(Duration(minutes: minutes)),
+    duration: Duration(minutes: minutes),
+    distance: 1000,
+    transitLeg: route != null,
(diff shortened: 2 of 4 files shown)


SamuelRioTz:
Verified on the real Sana'a app (Android 16 arm64 emulator, profile builds with `dependency_overrides` to `main` `61eac621` and to `a7ba58be`). Same sequence in both: deep link of the university trip → force-stop → cold start with data kept → screenshots at 3 s and 15 s.

**Root cause, confirmed with an instrumented build**: the restored route was already on the map — `TrufiMap` synced `route-lines-layer` (4 lines) and `route-markers-layer` (8 markers) right after the style loaded — but the camera never fitted it: the restore lands 3 ms after the first frame and 465 ms before the map's first camera report, `_fitCameraToPoints` returned on the null `_currentCamera`, and the one-shot replay in the first build had run 3 ms too early. The whole trip lies outside Sana'a's default viewport, hence the "empty" map (two swipes south on `main` reveal the "14" line, screenshot 4).

| `main` — after the deep link | `main` — cold start + 15 s (panel restored, map at the default centre) | `main` — same screen after panning south: the route was there, off-screen |
|---|---|---|
|  |  |  |

| this branch — after the deep link (fresh search, unchanged) | this branch — cold start + 3 s: restored route drawn and fitted | + 15 s (identical bytes) |
|---|---|---|
|  |  |  |

| rotated to landscape (wide layout): fit kept | back to portrait: kept | ✕ on the panel: route, markers and panel gone together | cold start after clearing: nothing restored |
|---|---|---|---|
|  |  |  |  |

**My-location after the restored fit** (found in the fresh review, fixed in `e7000192`): with the fit handed to the map as a controlled camera, a programmatic move (my-location) left `_cameraTarget` stale — the map suppresses its callback for its own moves, so the screen never released the fit and the next rebuild (every GPS tick) handed it back. Same sequence on both builds: cold start with the restored plan → GPS fix → tap my-location → 2 s → GPS tick → 5 s.

| `a7ba58be` — 2 s after my-location | `a7ba58be` — 5 s later, after a GPS tick: snapped back to the route | with the fix — 2 s after my-location | with the fix — 5 s later: stays on the user |
|---|---|---|---|
|  |  |  |  |

Review notes:

- The fix reuses the #946 replay instead of adding a second one: the fit is now computed against the initial camera while the map has not reported one (only the bearing comes from the base camera; `TrufiCameraPosition` has no tilt), and the controlled camera waits in `TrufiMap` for the style. The idle guard in `_handleCameraIdle` is what makes that wait survive Android's pre-style idle (`camera#onIdle` at 09.581, `onStyleLoaded` at 09.754 in the log) — without it the replay degenerates into an "already there" no-op.
- `flutter test` with Flutter 3.44.1: `trufi_core_home_screen` 39/39 (31 before), `trufi_core_maps` 56/56 (50 before). The 8 home-screen tests drive the real `HomeScreen` over a fake map engine (7 fail on `main`); the 6 map tests drive the real `TrufiMap` over a fake `MapLibrePlatform` via the plugin's `MapLibrePlatform.createInstance` seam (3 fail with `main`'s `trufi_map.dart`). Each guard has a mutation check: removing `_cameraTarget = null` fails exactly the my-location test; removing the idle guard fails exactly the three pre-style scenarios.
- No overlapping hunks with #991 (POI panel area of the same file).
- 0 crash lines in logcat across all runs; the fresh-search path takes the same code branch as before.

PR #999, opened 2026-09-12
Planner: honor TrufiPlannerConfig.maxTransfers and search two-transfer trips when nothing shorter exists (#998)

### LINKED ISSUE #998: Local planner: honor TrufiPlannerConfig.maxTransfers and support two-transfer trips (default stays 1)
## Problem

The on-device planner (`GtfsRoutingService.findRoutes`) enumerates direct trips and trips with exactly **one** transfer, and stops there (`gtfs_routing_service.dart:247-252`; in `_findOneTransferRoutes` the second pattern must already serve a destination stop, `:428-429`). Any trip that needs two changes of bus is unreachable by design. `TrufiPlannerConfig.maxTransfers` exists but is never passed to the routing client (`planner_routing_client.dart:15-20`): setting it in a city app changes nothing.

Reported as trufi-association/trufi-sanaa#2 (second reopening, 2026-09-10): جولة دار سلم (15.28064, 44.24430) → الحصبه هايبر ماركت (15.37844, 44.20752) returns "No routes" on v5.25.0 while OpenTripPlanner 1.5 on the same GTFS returns three itineraries, all with two transfers (7 → 14 → 14). A round-based search over the pattern graph confirms the minimum for that pair is two transfers and reproduces OTP's chain: 7 `19985848` → 8 m walk → 14 `18800916` → 73 m → 14 `19954455`.

How much it matters (400 random pairs ≥ 2 km, seed 42):

| Feed | reachable with ≤1 transfer (today) | with ≤2 |
|---|---|---|
| Sana'a (1500 m walk) | 54.0 % | 77.5 % |
| Cochabamba (800 m) | 95.8 % | 99.5 % |

Sana'a's network is fragmented (OSM-derived, 194 short lines, 157 of them numbered "7"); the UN-Habitat Sana'a City Profile notes that long north–south trips require changing buses several times.

## Proposal

Make the transfer limit real and configurable, default unchanged:

1. Wire `TrufiPlannerConfig.maxTransfers` through `PlannerRoutingClient` / `LocalPlannerClient` to `GtfsRoutingService.findRoutes`. **Default 1** (two buses) — every city keeps exactly today's results.
2. For `maxTransfers >= 2`, add a third phase: a round-based search over patterns using the existing #992 connection table (patterns as nodes, walkable/shared-stop connections as edges), bounded by `maxTransfers`, that runs **only when phases 1–2 return nothing** (so cities with the default see zero change, and cities that opt in only pay when it is needed). Score with the existing distance-based ranking plus the transfer penalty; emit the extra WALK legs between buses as #992 does. Measured cost of the round search: < 1 ms on Sana'a, 0.7 ms avg / 12 ms max on Cochabamba (desktop). The naive extension of the current pair loop to triples is not viable (up to ~200 M scans on Cochabamba).
3. trufi-sanaa sets `maxTransfers: 2` in its `TrufiPlannerConfig.local` (a separate app PR after the core release).

## Acceptance criteria

- Default (1): itinerary lists identical to v5.25.0 on the Cochabamba and Sana'a feeds for the 340/400 random-pair harnesses (byte-identical results).
- `maxTransfers: 2`: Ammar's pair returns the 7 → 14 → 14 itinerary (and the reverse trip); Sana'a reachability rises to ~77 % on the harness; `maxTransfers: 3` works too.
- UI: 3-leg itineraries render in the card, the detail and the #983 grouper; emulator evidence on the Sana'a app both ways (default 1 → "no routes"; 2 → itinerary).
- `TrufiPlannerConfig.maxTransfers` documented (default, cost, when to raise it); snapshot format of #994 unaffected (no new index).

## Related

- #992 walkable transfers (the connection table this search reuses); #994 persisted index; trufi-association/trufi-sanaa#2; OTP's own stance: it does not cap transfers, it searches by cost (`GraphPathFinder` overrides `maxTransfers`, opentripplanner/OpenTripPlanner#2522).

SamuelRioTz:
Closes #998. Engine side of trufi-association/trufi-sanaa#2 (second reopening).

## Motivation

The on-device planner enumerates direct trips and trips with exactly one transfer and stops there; `TrufiPlannerConfig.maxTransfers` exists but never reached the routing client, so setting it changed nothing. Any trip that needs two changes of bus is "No routes" by design. The reporter's trip جولة دار سلم → الحصبه هايبر ماركت (11.6 km, south → north) returns nothing on v5.25.0 while OpenTripPlanner 1.5 on the same GTFS returns three itineraries, all with two transfers (7 → 14 → 14). A round-based search over the pattern graph confirmed the minimum for that pair is two transfers and that no engine knob rescues it (2500 m walk, 150 m transfer radius, no candidate cap: still 0).

How much it matters, 400 random pairs ≥ 2 km per feed (seed 42): Sana'a plans 54.0 % of them today and 77.5 % with two transfers (90.3 % with three); Cochabamba 95.8 % → 99.5 %. Sana'a's network is 194 short OSM-derived lines (157 of them named "7"); the UN-Habitat Sana'a City Profile notes that long north–south trips require changing buses several times.

## Design

- **`maxTransfers` is real and configurable, default unchanged.** `TrufiPlannerConfig.maxTransfers` → `TrufiPlannerProvider` → `PlannerRoutingClient.findRoutes(maxTransfers:)` → `LocalPlannerClient` → `GtfsRoutingService.findRoutes`. Default **1** (two buses) at every layer: a city that does nothing keeps exactly today's results. `0` returns direct lines only; a negative value is rejected (`ArgumentError` in the service, `assert` in the config). `RemotePlannerClient` forwards it in the `/plan` body (a server on an older planner ignores it). No isolate change: planning runs on the root isolate; only the load does not.
- **Phase 3, only when phases 1 and 2 return nothing.** For `maxTransfers >= 2`, `findRoutes` runs a round-based search over the pattern graph (the idea of RAPTOR — Delling, Pajor, Werneck 2015 — without the time axis) on the connection table #992 already builds: nodes are patterns, edges the precomputed connections (alight at `myStopIdx`, board at `otherStopIdx`, `walkMeters` in between). Round 0 boards at the origin stops; round k scans the connections after the entry of every label created in round k − 1; from round 2 on, a hop into a pattern that serves a destination stop is an itinerary, and the first round that yields any ends the search. Fewest transfers win outright, exactly as a direct line hides one-transfer options today — so a city that opts in gets today's answer wherever today's answer exists, and pays for the search only on queries that would otherwise say "no routes" (46 % of them on Sana'a, 4 % on Cochabamba).
- **Exact ranking, bounded work.** Each pattern keeps a Pareto front of labels (entry position, accumulated cost): boarding earlier reaches more stops, boarding later may be cheaper; a label is dropped when riding an earlier one down to its entry costs no more. Fronts live in parallel typed arrays sorted by entry — offsets (cost − cumulative distance at the entry) strictly descend along a front — so dominance is one binary search and one comparison, and each new label only scans the alight positions between its entry and the next label's: every connection of a pattern is examined once per round, not once per label (around a dense origin a pattern carries 20+ non-dominated boardings). The hop into a destination pattern is scored on the spot (a suffix minimum answers "cheapest stop to alight after position b" in one binary search) and kept **best per chain of pattern ids**. Alternatives survive only where their labels survive: two chains that reach the same intermediate pattern collapse into the cheaper one there (Pareto dominance, or the expansion range handing the connection to the cheaper boarding), so the list offers distinct last legs and Pareto-incomparable prefixes — not every (first, second) pair as phase 2's best-per-pair does — then deduped by line key as everything else. The top itinerary is the exact optimum (checked against an exhaustive enumeration of every chain-rule-legal two-transfer chain on 94 Sana'a and 15 Cochabamba pairs); positions 2–5 differ from the exhaustive top five in 45 / 94 and 2 / 15 pairs. A chain never boards a line it already rode ("7 → 14 → back onto the same 7" is a detour, not an option); when the chain rule vetoes the cheapest label, the nearest earlier boarding **of the same round** with a different history takes the connection — a label of an earlier round never does, since its hop would carry fewer transfers than the round (a one-transfer chain phase 2 pruned on purpose). The Pareto rule can drop a label whose history was the only one the chain rule allowed for a later connection; measured on Sana'a, Cochabamba and Lima this never removed a reachable pair (0 / 352 phase-3 queries), and the chain rule itself never changed a result on Sana'a's 400 pairs. `GtfsRoutingService.maxMultiTransferScans` (default 6 M, measured maximum 2.65 M on Cochabamba) bounds a pathological query; when it runs out the search answers with what it found. Scores use the existing `walk × 2 + transit` formula; results of different transfer counts are never mixed, so no per-transfer penalty is needed.
- **No new index.** Same CSR connection table, same line keys; the #994 snapshot format is untouched (`PlannerIndexCodec` unchanged, its fixture digest test still passes). `GtfsRoutingService.multiTransferSearches` counts how many queries reached phase 3 (diagnostics; the tests pin the "only on zero results" rule on it).
- **UI.** Nothing to change: the card paints one chip per leg in a horizontally scrolling strip and shows transfers = transit legs − 1; the detail and the map iterate legs; the #983 grouper compares N slots; `Itinerary.toJson/fromJson` (#996 restored plan) round-trips three bus legs. Verified by tests and in the emulator.

## Measured (desktop, Flutter 3.44.1)

Default (`maxTransfers` 1) vs `main` `f35425e1` with `common_dump.dart` (#992's harness, timing fields stripped): **byte-identical** — Sana'a 344 queries / 560 itineraries, sha256 `ce46d1ea…` = `ce46d1ea…`; Cochabamba 340 queries / 1 241 itineraries, `012aa7b2…` = `012aa7b2…`. `compare.py`: 344/344 and 340/340 identical queries, 0 keys lost, 0 scores changed.

400 random pairs per feed (seed 42, endpoints = stop ± 300 m, ≥ 2 km apart; the diagnosis sample), app parameters (pool 150, 5 results):

| Feed (walk) | `maxTransfers` 1 (today) | 2 | 3 | pairs that plan with 1 whose list changes with 2 / 3 |
|---|---|---|---|---|
| Sana'a (1500 m) | 216 = **54.0 %**, 612 rows, 0.84 ms avg / 5.9 max | 310 = **77.5 %**, 933 rows, 0.89 ms avg / 2.9 max | 361 = **90.3 %**, 1.02 ms avg / 2.8 max | **0 / 0** |
| Cochabamba, trufi-app feed (800 m) | 383 = **95.8 %**, 1 487 rows, 8.6 ms avg / 88 max | 398 = **99.5 %**, 1 561 rows, 9.1 ms avg / 140 max | 398 = 99.5 %, 11.1 ms avg / 286 max | **0 / 0** |

Phase 3 itself (whole query, on the pairs that reach it): Sana'a 184 pairs, **0.60 ms avg / 1.8 ms max** with 2; Cochabamba 17 pairs, **26 ms avg / 140 ms max** with 2 (76 / 286 with 3) — the same pairs take 8.6 ms avg / 88 ms max today for "no routes". The diagnosis's 0.7 ms / 12 ms figure for Cochabamba was a single-label search that does not rank by cost; the Pareto-exact search with alternatives costs more on the dense feed, and Cochabamba keeps the default anyway. Ammar's pair on the full feed with 2: forward **2 itineraries** — 7 `19985848` (board 17 m from the origin) → 8 m walk → 14 `18800916` → 73 m → 14 `19954455` → 136 m, 17.8 km, score 18 267, plus the same chain through 14 `18904090`; reverse **4 itineraries**, best 7 `20083001` → 99 m → 7/14 `19955711` → shared stop → 7 `19985849` (OTP's second chain, mirrored). On the device (Sana'a app, profile build): "Found 2 paths in 5 ms" / "Found 4 paths in 4 ms".

## Tests

`trufi_core_planner` **85 → 121**, `trufi_core_routing` unit **140 → 148**, `trufi_core_home_screen` **46 → 48** (Flutter 3.44.1 = CI). `flutter analyze` adds nothing (the 3 infos in routing/home_screen are pre-existing, untouched files); `dart format` on the service and the new tests (the provider keeps its pre-existing drift; the inserted line is formatter style).

- `multi_transfer_routes_test.dart` (20): three lines in a row → nothing with 1, A → B → C with 2 (stops, indices, resolved stop lists, score = transit), 3 gives the same answer, 0 = direct only, phase 3 idle when phases 1–2 answer (counter), directed patterns, negative rejected, scan budget 0 → empty; four lines → 2 not enough, 3 finds four legs, a larger limit stops at the first round that answers; walkable transfers: two chains offered, walk-free first, 15 m walks measured and boarding at the real stop, score = 2 × walk + transit; chain rule: no line twice; dominance: cheaper boarding wins, a later cheaper boarding coexists with an earlier one that alone reaches a stop in between. Fewest transfers across rounds: a three-transfer chain is not offered next to a two-transfer one; the service default runs no multi-transfer search. Chain-rule fallback: a vetoed label never falls back to an earlier round (case L — that fallback would have emitted the one-transfer chain phase 2 pruned), the earlier boarding of the same round takes the vetoed connection (case F), and dominance can drop the only chain-rule-legal history (case H, known limitation, with its control).
- `sanaa_two_transfers_test.dart` (16) on a new 35 KB cut of the real feed (11 routes: five "7", five "14", one "7/14" — above `sameNameRouteLimit`, so each is its own line as in the full feed; a six-route cut would have merged the "14"s into one line and hidden the chain): the reporter's coordinates verbatim → nothing with 1 both ways; with 2 forward reproduces OTP stop for stop (`1223812364` → `11789489571`, `588459651` → `588459648`, 81 m of transfer walks, 136 m to the destination, resolved stop lists and shapes per leg), the alternative through `18904090` ranked after, reverse 7 → 7/14 → 7 at the shared stop `504469874`, exactly two transfers everywhere, 3 = 2, `LocalPlannerClient` passes the limit; and the four #992 pairs on that fixture return identical lists with 1, 2 and 3 while the counter stays 0. **Exact optimum**: six pairs (four random over the cut, the reporter's both ways) against an exhaustive enumeration of every chain-rule-legal two-transfer chain — first itinerary, indices, stops, transfer walk and the ordered scores of the whole list; without best-per-chain the first score is 10–15 % worse on two of them, without the removal of dominated labels a dearer boarding takes the first row on two others.
- `trufi_planner_multi_transfer_test.dart` (8): three-segment `convertToItinerary` → WALK BUS WALK BUS BUS WALK, contiguous times, transfers 2; JSON round trip; grouper keeps three-slot itineraries and offers the last slot's options; config default/configurable/negative; **end to end** through `TrufiPlannerProvider.fetchPlan` with the fixture served as a mocked asset (worker isolate as in the app): default → no transit itinerary, 2 → the 7 → 14 → 14 itinerary with its 8 m and 73 m walks and per-leg geometry.
- `itinerary_card_three_buses_test.dart` (2): seven chips, six chevrons, "2" transfer badge at phone width; a grouped third slot paints both options.

## Evidence (trufi-sanaa app copy on this branch, Android 16 emulator, profile build; before = default 1, after = `maxTransfers: 2` in `TrufiPlannerConfig.local`)

| Default (1) — جولة دار سلم → الحصبه: **no routes** | Default — reverse: **no routes** | Default — university trip already works (7 → 14, 57 min, 833 m) |
|---|---|---|
|  |  |  |

| `maxTransfers: 2` — **7 → 0' → 14 → 1' → 14**, 1 h 2 min, 18.2 km, 234 m, 2 transfers, three legs on the map | strip scrolled: the third bus and both walks | detail, full chain: **Walk 0 min (8 m)** at شارع تعز, **Walk 1 min (73 m)** at شارع الستين |
|---|---|---|
|  |  |  |

| detail, end: 67 stops → Walk 1 min (136 m) → destination | reverse: **7 → 7/14 → 7**, 46 min, 2 transfers ("2 routes found") | university trip **unchanged** (7 → 14, 57 min, 833 m) |
|---|---|---|
|  |  |  |

Logcat: 0 `Unhandled Exception|FATAL|Null check` across all runs. Cochabamba was verified with the harness on trufi-app's real feed (its toolchain still does not produce an APK from a clean copy, see #992).

## Notes for the reviewer

- Phase 3 emits itineraries with ≥ 2 transfers only. A one-transfer chain that phase 2 pruned (destination-bbox rule) stays pruned: the diagnosis counted 3 Cochabamba and 10 Lima pairs in that situation; unchanged here. The chain-rule fallback is restricted to labels of the same round for exactly this reason — an earlier-round fallback would have let a one-transfer chain out of phase 3 (case L in the tests).
- Sana'a results now show "7 → 14 → 14" and "7 → 7/14 → 7" rows; the grouper merges the engine's alternatives into one row per main bus (the forward pair shows "1 route found" for 2 paths, the reverse "2 routes found" for 4). Labelling same-named lines is the open product question from #992.
- trufi-sanaa sets `maxTransfers: 2` in a follow-up app PR after the core release; trufi-server-planner (Cochabamba web) keeps its default when rebuilt.

## Follow-ups (not blockers)

1. **Alternatives by first line.** Chains that share the intermediate pattern collapse into the cheapest one, so positions 2–5 differ from the exhaustive top five in 45 / 94 Sana'a pairs (top-1 always exact). Options: dominance per first line key, or a second pass that fills positions 2–5 with the alternative first legs onto the same intermediate pattern — parity with phase 2's best-per-pair.
2. **Dominance vs chain rule (case H).** A dominated label can carry the only history the rule accepts for a later connection; 0 / 352 real phase-3 queries affected. Options: keep dominated labels as shadows only the fallback may consult, dominance per first line, or drop the chain rule (RAPTOR has none) and let dominance and score decide.
3. **`multiTransferTruncations`** next to `multiTransferSearches`: nothing records a query that exhausted `maxMultiTransferScans`.
4. **Off the root isolate** if a dense feed ever opts in: Cochabamba's worst pair goes 84 → 119 ms (desktop) with 2, 205 ms with 3.
5. **trufi-server-planner**: the `/plan` handler must read `maxTransfers` and pass it to `findRoutes` before the web planner can honour it; bumping the planner pin alone changes nothing.
6. Card chip shows `0'` for sub-minute walks (8 m) — pre-existing since #992, cosmetic (`<1'`).


SamuelRioTz:
## Review — reproduced on the real Sana'a app, verified both ways, default proven byte-identical

**Verdict: works as described.** The reporter's trip from trufi-association/trufi-sanaa#2 (second reopening) was reproduced first with this branch at the **default limit** on the real trufi-sanaa app (v1.1.0 + these packages, Android 16 emulator, profile build, `trufiapp://route?…` deep link), then with `maxTransfers: 2` in `TrufiPlannerConfig.local`.

- **Root cause, confirmed on v5.25.0 / `main`**: `_findOneTransferRoutes` requires the second pattern to serve a destination stop and there is no third phase; `TrufiPlannerConfig.maxTransfers` never reached `PlannerRoutingClient`. No knob rescued the pair (2500 m, radius 150 m, no candidate cap: still 0), and OTP 1.5 on the same GTFS answers with two-transfer chains because it does not cap transfers at all (opentripplanner/OpenTripPlanner#2522).
- Code: the knob is wired through every layer with default **1**; the new phase runs **only when phases 1 and 2 return nothing** and stops at the first round that reaches the destination, so a city that opts in gets today's answer wherever today's answer exists. Round-based search over the existing #992 connection table (no new index, #994 snapshot untouched): Pareto fronts per pattern in typed arrays, one scan per connection per round, best-per-chain destination candidates, chain rule with a same-round fallback, scan budget. A second, independent pass checked phase 3 against an exhaustive enumeration of every chain-rule-legal two-transfer chain (94 Sana'a, 15 Cochabamba and the Lima pairs that reach it), a BFS lower bound on transfers, 21 mutations and three synthetic cases: the first itinerary is the exact optimum on every pair, all 957 emitted itineraries are valid stop for stop and connection for connection, and reachability matches the enumeration on all three feeds. It found one real defect — the chain-rule fallback could reach back to a label of an earlier round and let a one-transfer itinerary out of phase 3 (never on the 1 200 real pairs, reproducible on a synthetic outbound/return pair) — fixed in `29257a21` and pinned in `5f82d35c`; the answers on the real pairs did not change. Alternatives in positions 2–5 are fewer than the exhaustive top five where chains share the intermediate pattern (45 / 94 Sana'a pairs) — documented, follow-up below. `flutter analyze` adds nothing, `dart format` on the touched service and tests.
- **Default byte-identical to `main`** (#992's `common_dump` harness, timing stripped): Sana'a 344 queries / 560 itineraries, Cochabamba 340 / 1 241 — equal sha256, 0 keys lost, 0 scores changed. 400 random pairs per feed with 1 / 2 / 3: Sana'a **54.0 % → 77.5 % → 90.3 %** plannable, Cochabamba 95.8 % → 99.5 %, and **every pair that plans with 1 returns the identical list with 2 and 3** (0 / 0 differences on both feeds). Phase 3 cost on the queries that reach it: Sana'a 0.65 ms avg / 4.5 ms max; Cochabamba 28 ms avg / 150 ms max (desktop, whole query; the same queries take 9 / 88 ms today to answer "no routes") — Cochabamba keeps the default.
- Ammar's pair on the full feed with 2: forward 7 `19985848` → 8 m → 14 `18800916` → 73 m → 14 `19954455` (OTP's first itinerary stop for stop) plus one alternative; reverse 7 → 7/14 → 7 (OTP's second, mirrored), four itineraries. On the device: "Found 2 paths in 5 ms" / "Found 4 paths in 4 ms".
- Tests: planner **85 → 121**, routing unit **140 → 148**, home_screen **46 → 48** (Flutter 3.44.1 = CI). New real-feed fixture (11 routes — a 6-route cut would have merged the "14"s into one line under `sameNameRouteLimit` and hidden the chain), six fixture pairs pinned to the exhaustive optimum (chain, indices, stops, transfer walk, ordered scores of the whole list — they fail without best-per-chain candidates or without dominated-label removal), synthetic cases for the same-round fallback, the chain rule and the fewest-transfers rule, end-to-end plumbing through `TrufiPlannerProvider.fetchPlan` with the fixture as a mocked asset, three-leg conversion + JSON round trip + grouper, three-bus card at phone width.

### Evidence (trufi-sanaa app, Android 16 emulator, profile; before = this branch at the default 1, after = `maxTransfers: 2`)

| Default — جولة دار سلم → الحصبه: **no routes** | Default — reverse: **no routes** | Default — university trip works (7 → 14, 57 min, 833 m) |
|---|---|---|
|  |  |  |

| `maxTransfers: 2` — **7 → 0' → 14 → 1' → 14**, 1 h 2 min, 18.2 km, 234 m, 2 transfers | strip scrolled: third bus and both walks | detail, full chain: **Walk 0 min (8 m)**, **Walk 1 min (73 m)** |
|---|---|---|
|  |  |  |

| detail, end: 67 stops → Walk 1 min (136 m) → destination | reverse: **7 → 7/14 → 7**, 46 min, 2 transfers | university trip **unchanged** |
|---|---|---|
|  |  |  |

Logcat clean across all runs (0 `Unhandled Exception|FATAL|Null check`). The grouper merges the engine's alternatives into one row per main bus (forward shows "1 route found" for 2 paths; reverse "2 routes found" for 4). Cochabamba was verified with the planner harness on trufi-app's real feed (its toolchain still does not produce an APK from a clean copy, see #992).

1. trufi-sanaa: set `maxTransfers: 2` in `TrufiPlannerConfig.local` after the core release (app PR); tell the reporter that two changes of bus cover 77.5 % of random pairs, not everything (12.8 % need three, 9.8 % are out of reach within 1 500 m).
2. Alternatives by first line: chains that share the intermediate pattern collapse into the cheapest one (positions 2–5 differ from the exhaustive top five in 45 / 94 Sana'a pairs; top-1 always exact). Dominance per first line key, or a second pass filling positions 2–5 with the alternative first legs, would give parity with phase 2's best-per-pair.
3. Dominance vs chain rule: a dominated label can carry the only history the rule accepts for a later connection (0 / 352 real phase-3 queries affected; pinned as a known limitation in the tests). Options: shadow labels the fallback may consult, dominance per first line, or drop the chain rule as RAPTOR does.
4. `multiTransferTruncations` next to `multiTransferSearches`: nothing records a query that exhausted `maxMultiTransferScans` (measured maximum 2.65 M of 6 M).
5. If a dense feed ever opts in, the round-1 scan is the remaining cost (~120 ms worst case on Cochabamba, desktop, 205 ms with 3); `compute()` for phase 3, or a reverse adjacency for the last round (a new index column), would address it.
6. trufi-server-planner: the `/plan` handler must read `maxTransfers` and pass it to `findRoutes` before the web planner can honour it; bumping the pin alone changes nothing.
7. Labelling same-named lines ("7 → 14 → 14", "7 → 7/14 → 7") — the product question from #992, heavier with three legs.
8. The card chip shows `0'` for sub-minute walks (8 m) — pre-existing since #992, cosmetic (`<1'`).

Phase 3 emits ≥ 2 transfers only: a one-transfer chain that phase 2's destination-bbox rule prunes stays pruned (3 Cochabamba / 10 Lima pairs in the diagnosis), and the same-round fallback is what keeps that guarantee.

PR #1000, opened 2026-09-16
Planner: take ride durations from stop_times and make the fallback speed configurable (#997)

### LINKED ISSUE #997: Local planner: bus speed is hard-coded at 18 km/h; align it with the feed or make it configurable
## Problem

`TrufiPlannerProvider` (the offline planner on mobile, remote on web — first engine in trufi-app's list) estimates in-vehicle time from distance at a **hard-coded 18 km/h** and ignores the feed's `stop_times`:

```dart
// packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart (main, v5.25.0)
// Estimate in-vehicle time from transit distance at ~18 km/h average
// urban transit speed. Independent of stop spacing — works for both
// sparse feeds (stops every ~500m) and dense feeds (stops every ~30m).
final estDuration = Duration(seconds: (transit / 5).round());
```

The feed itself carries a different figure. trufi-gtfs-builder writes `stop_times` from `distance / vehicleSpeed`; Cochabamba's config had 40 km/h and, after trufi-association/trufi-gtfs-builder#9, moves to **20 km/h for buses**, its own OSM `duration=*` for Mi Tren, and route-type-specific values (trufi-association/trufi-gtfs-builder#16). OTP 2.8.1 (the alternative engine in Settings, and planner.trufi.app's API) and Transitous/MOTIS read those `stop_times`; the Trufi Planner does not.

So **the same trip gets two different durations depending on the engine**: micro H end to end (19 km) is 30 min in the feed at 40 km/h, ~59 min at 20 km/h, and 64 min in the Trufi Planner at 18 km/h — and a user switching engines sees the numbers jump. The 18 also has no source: it is a constant chosen in #890 (May 2026) "independent of stop spacing", which was the right fix for the stop-density problem but leaves the speed itself unexplained. The reporter of gtfs-builder#9 saw "17 minutes" for a ride that took an hour; 17 min is what the Trufi Planner gives for ~5 km at 18 km/h.

## Proposal

Either of these — the first is the smaller change, the second removes the duplication:

1. **Make the speed configurable and default it to the feed's figure.** Add `averageSpeedKmh` (or a per-`route_type` map: bus 20, light rail 32 in Cochabamba) to `TrufiPlannerConfig.local` and `.remote`, default **20 km/h**, and use it in `_buildSegmentForPattern` instead of `transit / 5`. City apps that publish a feed at another speed set the same number in both places. Include it in the persisted-index key only if it affects the index (it should not: durations are computed per segment at query time).
2. **Read the duration from `stop_times` when the feed has real timings.** With the builder now honoring OSM `duration=*` (and, later, measured running times from Cochabamba's GTFS-RT vehicle positions), `arrival_time` differences between the boarding and alighting stops are the truth for those routes. The parser already reads `stop_times.txt`; `RoutePattern` would need the cumulative offsets alongside `cumDist`, and `_buildSegmentForPattern` would use `offset[toIdx] - offset[fromIdx]` when available, falling back to the speed estimate otherwise. Frequency-based feeds (Cochabamba: `exact_times=1`, headway 300 s) express `stop_times` as offsets from `00:00:00`, so the subtraction works directly.

Either way the planner and the feed should agree on one figure per route type, documented next to `walkSpeed` in `TrufiPlannerConfig`.

## Context

- Builder side: trufi-association/trufi-gtfs-builder#9 (the 40 km/h question) and the PR above (20 km/h buses, OSM `duration=*`, per-route-type callback).
- Where the 18 km/h lives: `packages/trufi_core_planner/lib/src/services/gtfs_routing_service.dart`, `_buildSegmentForPattern` (`transit / 5`), introduced in #890.
- Public reference for Cochabamba bus speeds: municipal Dirección de Tráfico y Vialidad, 2017 — 10–11 km/h in congestion, 25 km/h "satisfactory" (https://ibce.org.bo/principales-noticias-bolivia/noticias-nacionales-detalle.php?id=74617&idPeriodico=7&fecha=2017-03-09).
- Affected apps: every city app using `TrufiPlannerConfig.local` (Cochabamba, Sana'a, …); the change is a default plus an option, so no app breaks.

SamuelRioTz:
Closes #997.

## Motivation

`TrufiPlannerProvider` (the offline planner on mobile, `trufi-server-planner` on the web) estimated every ride from its distance at a hard-coded 18 km/h (`transit / 5`, introduced in #890) and never read the feed's `stop_times`. OTP 2.8.1 and MOTIS/Transitous do read them, so the same trip had two durations depending on the engine, and the 18 itself had no source. The reporter of trufi-association/trufi-gtfs-builder#9 saw "17 minutes" for an hour-long ride: that number came from this constant, not from the feed.

## Design

- **`RoutePattern` carries the timings of the trip that defined it** — the first trip, in `stop_times` order, with that stop sequence (builder feeds mostly have one trip per pattern; Lima has three identical ones; a timetable's first trip stands for the rest). `arrivalOffsets[i]` / `departureOffsets[i]` are seconds from the first arrival, read from `stop_times` while the pattern is built; `scheduledSecondsBetween(from, to)` is the arrival at the alighting stop minus the departure from the boarding stop (dwell at the boarding stop is not ride time). Frequency-based feeds write `stop_times` as offsets from `00:00:00`, timetables as clock times — only differences are used, so both work.
- **Timings are dropped, never guessed.** A stop with neither `arrival_time` nor `departure_time`, times that run backwards, or a row count that does not match the stop list leave the pattern without timings; a blank arrival takes the row's departure and vice versa (GTFS allows that for timepoints); consecutive rows for the same stop collapse exactly like the stop list already does (first arrival, last departure). A ride whose scheduled time is 0 (a feed with every time `00:00:00`) is treated as unknown.
- **`GtfsRoutingService.fallbackVehicleSpeedKmh` (default 20, must be > 0)** is used only for rides on patterns without timings. It reaches the service through `LocalPlannerClient` from the new `TrufiPlannerConfig.local(fallbackVehicleSpeedKmh:)`; remote mode leaves it to the server (`trufi-server-planner` builds its own `GtfsRoutingService` and picks the new behaviour up when it bumps its pin). 20 km/h is what the issue settled on and what trufi-app#49 (open) sets for Cochabamba's buses in its builder config (trufi-gtfs-builder#9/#16 left the library default alone); today's published feed is still at 40 km/h, which is why the branch reads ~38 km/h on it below.
- **Ranking is untouched.** The score is walk × reluctance + transit distance; durations are display data. Verified below: the itinerary lists are identical on three real feeds, only the minutes change.
- **Snapshot (#993).** The offsets are persisted next to `cumDist` (a flag per pattern, then the timed patterns' offsets back to back as adaptive-width unsigned columns); `PlannerIndexCodec.formatVersion` goes 1 → 2, so every cached index is rebuilt once on the first start after the update, and the format-guard digest is updated.

## Measured (desktop, Flutter 3.44.1; 344–345 queries per feed: the #999 random set, seeds 42/300+7/40, plus end-to-end rides and Ammar's Sana'a pairs)

| Feed (builder speed in its `stop_times`) | itinerary lists identical to `main` | legs whose duration changed | duration ratio branch / main (median) | implied speed before → after | build ms main / branch |
|---|---|---|---|---|---|
| Cochabamba, trufi-app `main` feed (40 km/h) | **344 / 344** | 1 994 / 1 994 | 0.47 | 18.0 → 38.3 km/h | 532 / 531 |
| Cochabamba, trufi-app PR #49 feed (20 km/h buses, OSM `duration=*` for Mi Tren) | **344 / 344** | 1 993 / 1 994 | 0.92 | 18.0 → 19.6 km/h (18–48: Mi Tren keeps its own timings) | 531 / 557 |
| Sana'a, trufi-sanaa feed (30 km/h) | **345 / 345** | 909 / 921 | 0.62 | 18.0 → 28.9 km/h | 63 / 64 |

Micro H end to end (the reporter's ride, 17.5 km between its first and last stop): 58.5 min on `main` → **27.4 min** with today's 40 km/h feed, **53.8 min** with the PR #49 feed — the same figures OTP shows for that feed. Sana'a's rides get ~40 % shorter for the same reason: the feed says 30 km/h; the app used to say 18. Average query time unchanged (8.2 → 8.3 ms Cochabamba, 1.05 → 1.03 ms Sana'a).

Implied speed is computed over the straight-line stop-to-stop distance the planner uses, so it reads a little below the builder's shape-based speed.

## Tests

- `trufi_core_planner`: **121 → 162** (new `pattern_stop_times_test.dart`: clock times with dwell, frequency offsets, out-of-range indices, duplicate consecutive rows, blank cells, backwards times, all-zero times, JSON patterns, first trip wins, feed timing used for a ride, default 20 km/h, configurable fallback only touching untimed rides, non-positive speed rejected, `LocalPlannerClient` wiring; the snapshot round trip now checks both offset columns on a fixture whose patterns are all timed; format-guard digest updated with `formatVersion` 2; and, from the fresh review, `stop_times_edge_cases_test.dart` with 26 edge cases: circular lines, which trip defines a pattern, times past 24:00:00, stops missing from `stops.txt`, one-stop patterns, blank first cells, snapshot round trip with mixed timed/untimed patterns plus corrupted flags and truncation, the knob through `loadFromBytes`).
- `trufi_core_routing`: **152 → 155** (default 20, custom value, assertion on 0, remote fixed).
- Analyzer clean on both packages (the two `use_null_aware_elements` infos in routing are pre-existing on `main`); `dart format` clean on every touched file.

## Follow-ups (not in this PR)

- `trufi-server-planner` pins core v5.24.0; once it bumps, planner.trufi.app's API returns the same feed-based durations.
- Sana'a's app will show shorter rides after its next core bump; worth a sentence to Ammar when that happens.


SamuelRioTz:
## Review — ride durations now come from the feed; ranking proven byte-identical on three real feeds

**Verdict: works as described.** The planner used to estimate every ride at a hard-coded 18 km/h; with this branch a ride's duration is the feed's own `stop_times` figure (arrival at the alighting stop minus departure from the boarding stop), and only patterns without usable timings fall back to `fallbackVehicleSpeedKmh` (default 20, configurable in `TrufiPlannerConfig.local`).

- **Ranking untouched, measured**: the author's harness re-run independently (344–345 queries per feed, the #999 random set plus end-to-end rides) gives itinerary lists — keys, board/alight indices and scores — identical to `main` on Cochabamba (today's feed and the PR #49 feed) and Sana'a; only the minutes change (median ratio 0.47 / 0.92 / 0.62). Micro H end to end: 58.5 min → 27.4 (today's 40 km/h feed) / 53.8 (PR #49 feed), the figures OTP gives for the same feeds.
- **Timings read conservatively**: a stop without both times, backwards times or a row-count mismatch leave the pattern untimed → fallback; consecutive duplicate rows collapse like the stop list; an all-`00:00:00` trip counts as unknown. A probe over 8 feeds (Cochabamba ×2, Sana'a, Lima, Trujillo, a timetable feed with 8 709 trips, 2 fixtures) found 0 violations and 100 % of patterns timed, and the offsets equal an independent recomputation from `GtfsData.stopTimes`.
- **Snapshot**: offsets persisted, `formatVersion` 1 → 2 (one rebuild on the first start after the update, 1.3 s on the example feed), +1.1 % size on Cochabamba; corrupted flags / truncation decode to `null`, never throw.
- **Fresh review (CAMBIOS MENORES) applied in `ec5a96d9`**: 26 edge-case tests folded into the suite (circular lines, which trip defines a pattern, times past 24:00:00, missing stops, one-stop patterns, blank first cells, mixed timed/untimed round trip, corrupted snapshot, `loadFromBytes` wiring), the "one trip per pattern" comment corrected (Lima carries three), config defaults to the planner's constant. Mutation survivors m07/m21/m23–m28 are now caught.
- Tests: planner **121 → 162**, routing **152 → 155** (Flutter 3.44.1 = CI); analyzer clean; `apps/example` web build compiles; CI green on both commits.

### Evidence (`apps/example`, Android 16 emulator; Av. Chapare → Av. Río Parapeti, micro H end to end; the example's bundled feed says 00:34:27 for that ride)

| `main`: **1 h 0 min** (18 km/h) | branch: **34 min** = the feed | detail: 34 min, 19.6 km, 40 stops |
|---|---|---|
|  |  |  |

| OTP 2.8.1 in the same app: 29 min (the server runs trufi-app's newer feed at 40 km/h, so the remaining gap is feed, not engine) | feed with every `arrival_time`/`departure_time` blanked → fallback **54 min** (17.5 km at 20 km/h) |
|---|---|
|  |  |

First start with the branch over a `main` install: `cache rejected: format version 1, expected 2` → index rebuilt in 1.3 s and rewritten. Logcat clean (0 `Unhandled Exception|FATAL|Null check`) across all runs.

### Follow-ups (not blocking)

- Sana'a's rides get ~40 % shorter on the next core bump (its feed says 30 km/h; the app said 18) — worth a line to the reporter of trufi-sanaa#2.
- `trufi-server-planner` pins v5.24.0; planner.trufi.app's API picks the feed-based durations up when it bumps.
- Agency feeds with blank cells on non-timepoints (allowed by GTFS) fall back per pattern instead of interpolating; no Trufi feed has them.

PR #1001, opened 2026-09-16
Search: show the street's municipality under offline street and corner rows (#972)

### LINKED ISSUE #972: Street rows in search results carry no locality — same-named streets from different towns are indistinguishable
Found while testing the corner search on the Cochabamba app (core v5.21.0).

**Case**: searching `ayacucho` returns several street rows — among them two distinct "Calle Ayacucho" from different municipalities inside the PBF extract. The rows show only the street name, so they're indistinguishable; picking the wrong one silently set a corner **30 km east** of the city (lon -65.72 vs Cochabamba's -66.15), and the subsequent OTP plan legitimately returned no routes (out of transit coverage).

Place rows (Photon) already show context ("Sacaba, Cochabamba, Bolivia"); offline street/corner rows show none.

**Proposal**: carry a locality per street in the offline index and render it as the row subtitle (and in the corners screen title).
- Index side: `osm-search-data-export` could resolve each street's admin area at build time (the PBF has the boundaries), or approximate by nearest place node.
- App side: `OfflineSearchDataService` surfaces it via the existing `address`/subtitle field — the search UI already renders subtitles for Photon results.

Until then, corner picks outside the city are a silent footgun for any extract broader than one town.

SamuelRioTz:
Closes #972.

## Motivation

Offline street and corner rows show only the street name. In an extract wider than one town — Cochabamba's covers the whole metropolitan area plus the Valle Alto — two "Calle Ayacucho" or two "Calle Sucre" from different municipalities look identical, and picking the wrong one silently plans from a point 30 km away (the OTP answer is then a legitimate "no routes"). Photon rows already carry their context ("Sacaba, Cochabamba, Bolivia") as a subtitle; the offline rows did not.

## What changes

- `OfflineSearchDataService` reads the fourth element of each `streets` row of the exporter's `json-compact` index, `region` — the municipality (OSM `admin_level=8` boundary) the street centre falls in — and sets it as the result's `address`. The search screen already renders `address` as the row subtitle, and the origin/destination field already shows "name, address", so the UI needs no new strings.
- Corners: a corner's `address` is the municipality of the street it belongs to; when the other street lies in a different municipality it reads `"Cercado / Sacaba"` (a boundary corner).
- The corners sub-screen's header shows the municipality under the street name; corner rows in that list only add their own when it differs from the header's, so the list is not a column of identical subtitles.
- **One street entry per municipality is assumed.** The corner subtitle is derived from the two streets' regions (the index carries no region per corner), so it is only right when every street entry belongs to a single municipality. `osm-search-data-export` did the opposite until now — it merged every same-named way of the extract into one entry with one centre — which is why a regionalised index from the current exporter labels downtown corners "Cochabamba / Punata"; the exporter fix (regions from PBF boundaries **and** streets split by name + municipality) lands first, and only then is the trufi-app index regenerated. The doc comment states the premise.
- **No-op for today's indexes.** A row with three elements, or a `region` that is `null` or empty, produces exactly the results it produces on `main`: no subtitle, nothing else changes. Cochabamba's current index has `region: null` on all 4 025 streets because of a bug in `osm-search-data-export` with PBF input (fixed in trufi-association/osm-search-data-export — separate PR; the trufi-app index is regenerated in its own PR once that lands). This PR is safe to merge before either.

## Tests (Flutter 3.44.1 = CI)

`trufi_core_search_locations`: **142 → 157**. Unit (`offline_search_data_service_test.dart`, new group "street locality"): region present, three-element row, empty string, JSON `null`, same-region corner, boundary corner, one side unknown (both directions), both unknown, `junctionsOf`. Widget (`location_search_screen_drilldown_test.dart`): subtitle under a street result, header line under the street name, bare header without a region, corner rows not repeating the header's municipality while a boundary corner shows both; from the fresh review: padded region trimmed, blank and non-string regions ignored without sinking the index, empty-string header stays bare. Analyzer clean.


SamuelRioTz:
## Review — municipality under offline street and corner rows; a no-op until the index carries it

**Verdict: works as described.** Offline street rows and corners now show the municipality the exporter puts in each street's `region` field, through the `address` channel the screen already renders for Photon rows; the corners sub-screen names it in its header and corner rows only add their own when it differs (a boundary corner reads "Colcapirhua / Quillacollo").

- **No-op proven on today's index**: with trufi-app's real `search.json` (4 025 streets, `region: null` everywhere) a full sweep — every street name as a query plus the drill-down of all 3 962 drillable streets, 35 956 result lines — is byte-identical between `main` and this branch (sha256 equal). With a regionalised index the same sweep differs only in the `address` column (27 968 / 27 968 rows; 0 id/name/coordinate changes).
- **Malformed values**: numeric, list, map, boolean, padded or blank `region` values are ignored without sinking the index; an RTL region renders start-aligned; a 91-character region stays one line with an ellipsis; the region is not searchable; saved places and the composite dedupe compare by id/name/coordinates, so a stored place without `address` and the same one with it are one place.
- **Premise stated**: a corner's subtitle is derived from its two streets' regions, so it is only right when each street entry belongs to one municipality. `osm-search-data-export` merged same-named ways of the whole extract into one entry until now (that produced "Cochabamba / Punata" corners downtown with a regionalised index — screenshot 03 below); trufi-association/osm-search-data-export#22 fixes the region resolution **and** splits streets by name + municipality, and the trufi-app index is regenerated only after it. The doc comment says so.
- **Fresh review (CAMBIOS MENORES) applied in `3a7e1bd1`**: the premise documented, three tests added (padded region trimmed, non-string regions ignored, empty-string header). Mutation survivors closed.
- Tests: `trufi_core_search_locations` **142 → 157** (Flutter 3.44.1 = CI); analyzer clean; CI green on every commit.

### Evidence (`apps/example`, Android 16 emulator, search "ayacucho")

| today's index (`region: null`): identical to `main`, no subtitle | regionalised index from the fixed exporter, before the split: "CalleAyacucho — Colomi" (the 30 km footgun) vs "Calle Ayacucho — Cochabamba" | …but its corners exposed the exporter's name-merging: "& Calle Abaroa — Cochabamba / Punata" |
|---|---|---|
|  |  |  |

| index from osm-search-data-export#22 (split by municipality): one "Calle Ayacucho" per town | corners of the Quillacollo one: 9, all in Quillacollo, header carries the municipality | "Avenida Blanco Galindo" per municipality |
|---|---|---|
|  |  |  |

| corners of the Colcapirhua one: only the corner at the limit, where the avenue changes name, reads "Colcapirhua / Quillacollo" |
|---|
|  |

Logcat clean (0 `Unhandled Exception|FATAL|Null check`) across all runs.

### Follow-ups (not blocking)

- trufi-app: bump `osm-search-data-export` in `tools/search-bolivia-cochabamba` once #22 is tagged and regenerate `assets/search/search.json` (4 025 → 4 434 streets).
- A same-named street continuing across a municipal limit is no longer emitted as a corner by the exporter (178 such rows in Cochabamba dropped in osm-search-data-export#22 after its own review), so "Calle X & Calle X" rows never reach this screen.

